PR #5, opened 2021-01-26
[DBZ-2974]: Fix deserialization of List<UserType> columns

bingqinzhou:
https://issues.redhat.com/browse/DBZ-2974

### DIFF
--- a/src/main/java/io/debezium/connector/cassandra/CellData.java
+++ b/src/main/java/io/debezium/connector/cassandra/CellData.java
@@ -7,17 +7,17 @@ package io.debezium.connector.cassandra;
 
 import java.util.Objects;
 
-import io.debezium.DebeziumException;
 import org.apache.cassandra.db.marshal.AbstractType;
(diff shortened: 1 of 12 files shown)


bingqinzhou:
@gunnarmorling Can I get a review for this PR when you get a chance please? Thank you!

gunnarmorling (approved):
LGTM, but could you add a test for the affected situation, too? Thx!

bingqinzhou:
> LGTM, but could you add a test for the affected situation, too? Thx!

Added tests for this :)

gunnarmorling:
Excellent, thx! If I could suggest one thing, it'd be to squash the commits into one in cases like this, where there are many intermediary commits, which aren't interesting in the long term to keep separate. But no worries, we can do that next time. Thanks again!

gunnarmorling:
Debezium 1.5.0.Alpha1 (including this fix) is scheduled for next week btw.

PR #9, opened 2021-02-06
[DBZ-3063]: Fix CommitLogTransfer Config.

bingqinzhou:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,13 +4,13 @@
     <parent>
         <groupId>io.debezium</groupId>
         <artifactId>debezium-parent</artifactId>
-        <version>1.5.0-SNAPSHOT</version>
+        <version>1.4.0.Final</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
     <artifactId>debezium-connector-cassandra</artifactId>
     <name>Debezium Connector for Cassandra</name>
-    <version>1.5.0-SNAPSHOT</version>
+    <version>1.4.0.Final</version>
     <packaging>jar</packaging>
 
     <scm>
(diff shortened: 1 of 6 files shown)


ahmedjami:
Hi @bingqinzhou 
According to the  [connector properties](https://debezium.io/documentation/reference/connectors/cassandra.html#cassandra-connector-properties) table, I can't see any property related to this PR (COMMIT_LOG_TRANSFER_CONFIG_PREFIX).
Could you give me more details please, So i can add this to docs if necessary ?
Thanks in advance :)

gunnarmorling:
Good catch, @ahmedjami!

I think I remember some details around this; when WePay open-sourced this connector (via @jgao54 and @criccomini), we had to cut some ties to internal infrastructure/services which aren't available to other users of this connector. {{CommitLogTransfer}} is a service interface resulting from this split, which can be used to to copy Cassandra logs to an external location (e.g. on S3 or similar) for archiving purposes, so they can be removed on the actual Cassandra nodes after processing them.

IIUC, the properties with that prefix discussed here are passed through to the specific {{CommitLogTransfer}} implementation. Some update of the docs is warranted, e.g. there's still mention of [GCSCommitLogTransfer](https://debezium.io/documentation/reference/connectors/cassandra.html#cassandra-property-commit-log-transfer-class), which doesn't exist in Debezium (I suppose WePay has this as an internal implementation now). I.e. the `commit.log.transfer.class`, `remote.commit.log.relocation.dir`and `storage.credential.key.file` properties could need some doc update.

@bingqinzhou, can you clarify the notion of "commitLogRelocationDir" a bit further; the docs describe this as a _remote_ directory, whereas in the code it's handled as a local directory on the node's file system. I.e. it's a bit unclear to me right now how these things are meant to work. Thanks!

bingqinzhou:
Closing this one since the commit history is kind of messed up, created https://github.com/debezium/debezium-connector-cassandra/pull/11 instead.

PR #10, opened 2021-02-08
DBZ-3060 Add support for cassandra version >=3.11.5

ahmedjami:
https://issues.redhat.com/browse/DBZ-3060

Test OK for Cassandra 3.11.3, 3.11.4 .... And the latest version 3.11.10

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,7 @@
         <dependency>
             <groupId>org.apache.cassandra</groupId>
             <artifactId>cassandra-all</artifactId>
-            <version>3.11.4</version>
+            <version>3.11.10</version>
             <exclusions>
                 <exclusion>
                     <groupId>ch.qos.logback</groupId>


gunnarmorling:
Ha, thought for a second that'd add support for Cassandra 4.x :) That being said, this solely updates the Cassandra version being used; no further code changes within Debezium itself are needed?

ahmedjami:
According to tests that I realized, no code changes are needed for DBZ. No major changes on CommitLog classes for Cassandra 3.11.X series.

It will be the case when we will add support for Cassandra 4.x :)

gunnarmorling:
Ok, cool, going to apply it then. Speaking of Cassandra 4.x, do you have any feeling what that task would entail?

ahmedjami:
I will send an email to discuss support of Cassandra 4.x :)

gunnarmorling:
Awesome, thanks a lot! Best place would be the mailing list here: https://groups.google.com/g/debezium

bingqinzhou:
@ahmedjami As Cassandra 4.0 is not out yet, what's the purpose for upgrading cassandra-all version at this moment? I'm concerned that this version might break people who are currently running older version of Cassandra, or this new version might affect some current functions (corresponding with the old cassandra-all version) in debezium cassandra connector. Has these ^ been tested and confirmed ok yet? Since we're still testing cassandra connector and try to make it as stable as possible, we don't won't deal with any new bug brought by a new version of cassandra-all lib at this moment. cc: @gunnarmorling

gunnarmorling:
Hey @bingqinzhou, leaving it to @ahmedjami to comment on the details, but he tested "...for Cassandra 3.11.3, 3.11.4 .... And the latest version 3.11.10". I'd also expect such micro updates to generally be fully compatible drop-in replacements? Or is Cassandra pursuing another kind of versioning scheme? We also can add further versions for testing in a matrix job on CI; if that's useful, can you log a Jira issue stating the versions that should be tested?

bingqinzhou:
@gunnarmorling Thanks for the explanation:) We're currently using Cassandra 3.11.9 so it should be covered by the test already. Since CI build passed, I suppose everything should be compiling and working as before. @ahmedjami Please ignore my previous comment and thanks for fixing the issue!

ahmedjami:
@bingqinzhou np ;)

PR #11, opened 2021-02-10
[DBZ-3063]: Modify Cassandra Connector to work with customized CommitLogTransfer smoothly

bingqinzhou:
…ommit log transfer configs.

DBZ-3063 Change sleep time for CommitLogPostProcessor.

DBZ-3063 Only download error commitLog files upon starting.

DBZ-3063 Modify error logs when deleting commitlog files.

### DIFF
--- a/src/main/java/io/debezium/connector/cassandra/CommitLogPostProcessor.java
+++ b/src/main/java/io/debezium/connector/cassandra/CommitLogPostProcessor.java
@@ -20,7 +20,6 @@ import java.util.concurrent.TimeUnit;
 public class CommitLogPostProcessor extends AbstractProcessor {
 
     private static final String NAME = "Commit Log Post-Processor";
-    private static final int SLEEP_MS = 100;
     private static final int THREAD_POOL_SIZE = 10;
     private static final int TERMINATION_WAIT_TIME_SECONDS = 10;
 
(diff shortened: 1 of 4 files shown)


bingqinzhou:
@gunnarmorling May I get a review for this PR when you get a chance please? Thank you!

src/main/java/io/debezium/connector/cassandra/CommitLogPostProcessor.java
```diff
@@ -20,7 +20,6 @@
 public class CommitLogPostProcessor extends AbstractProcessor {
 
     private static final String NAME = "Commit Log Post-Processor";
-    private static final int SLEEP_MS = 100;
```
gunnarmorling, line 23: So the default value of this changed (now 10_000 ms), I suppose it's intentionally?

gunnarmorling (approved):
One question inline, but I think we're good to go with this. Thx!

src/main/java/io/debezium/connector/cassandra/CommitLogPostProcessor.java
bingqinzhou: > So the default value of this changed (now 10_000 ms), I suppose it's intentionally?

Yes, otherwise CommitLogPostProcessor will keep submitting upload job repeatedly for the same filed while it is being uploaded.

PR #12, opened 2021-02-12
DBZ-3063 Using Duration more consistently

gunnarmorling:

@bingqinzhou, quick follow-up to your PR; could you review and merge if it's ok? Thx!


bingqinzhou:
> https://issues.redhat.com/browse/DBZ-3063
> 
> @bingqinzhou, quick follow-up to your PR; could you review and merge if it's ok? Thx!

@gunnarmorling May I understand what the `Duration` related changes are for please?

gunnarmorling:
We generally try to use strong types such as Instant, Duration, etc. instead of long-encoded milli-seconds, so to make clear what the values are. Then it's also not needed to encode the unit into variable names ("...ms"). So the PR is some clean-up essentially :)

PR #15, opened 2021-03-06
[DBZ-3224] Improve error handling logic when failing to publish change events.

bingqinzhou:
https://issues.redhat.com/browse/DBZ-3224

### DIFF
--- a/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
+++ b/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
@@ -6,6 +6,7 @@
 package io.debezium.connector.cassandra;
 
 import java.util.Collections;
+import java.util.HashSet;
 
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.config.Schema;
@@ -23,6 +24,7 @@ public class CassandraConnectorContext extends CdcSourceTaskContext {
     private final ChangeEventQueue<Event> queue;
     private final SchemaHolder schemaHolder;
     private final OffsetWriter offsetWriter;
+    private final HashSet<String> errorCommitLogSet;
 
     public CassandraConnectorContext(CassandraConnectorConfig config) throws Exception {
 
@@ -30,6 +32,10 @@ public class CassandraConnectorContext extends CdcSourceTaskContext {
         this.config = config;
 
         try {
+
+            // Create a HashSet to record names of CommitLog Files which are not successfully read or streamed.
+            this.errorCommitLogSet = new HashSet<>();
+
             // Loading up DDL schemas from disk
             loadDdlFromDisk(this.config.cassandraConfig());
 
@@ -98,5 +104,9 @@ public class CassandraConnectorContext extends CdcSourceTaskContext {
 
     public SchemaHolder getSchemaHolder() {
         return schemaHolder;
+    }
+
+    public HashSet<String> getErrorCommitLogSet() {
+        return errorCommitLogSet;
     }
 }
--- a/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
+++ b/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
@@ -7,6 +7,7 @@ package io.debezium.connector.cassandra;
 
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Set;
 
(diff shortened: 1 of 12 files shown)


src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
```diff
@@ -23,13 +24,18 @@
     private final ChangeEventQueue<Event> queue;
     private final SchemaHolder schemaHolder;
     private final OffsetWriter offsetWriter;
+    private final HashSet<String> errorCommitLogSet;
```
gunnarmorling, line 27: Let's just expose `Set` here and as the return value of the method (here and elsewhere in this change set). As a name perhaps just `erroneousCommitLogs`, or similar? No need to encode the type into the variable name.

src/main/java/io/debezium/connector/cassandra/KafkaRecordEmitter.java
```diff
@@ -41,28 +43,35 @@
...
+                ProducerRecord<byte[], byte[]> producerRecord = toProducerRecord(record);
                 Future<RecordMetadata> future = producer.send(producerRecord);
                 LOGGER.debug("The record '{}' has been sent", record.toString());
                 futures.put(record, future);
                 maybeFlushAndMarkOffset();
             }
         }
         catch (Exception e) {
-            throw new DebeziumException(String.format("Failed to send record %s", record.toString()), e);
+            LOGGER.error("Failed to send the record {}. Error: ", record.toString(), e);
```
gunnarmorling, line 70: I'd suggest to _either_ log _or_ rethrow, so not to have the same exception in the logs twice.

src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
gunnarmorling: Is this set accessed from multiple threads at the same time btw.?

gunnarmorling (commented):
Hey @bingqinzhou, thanks a lot! A few comments inline.

gunnarmorling:
Looks like there's a compilation failure on CI.

src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
bingqinzhou: > Let's just expose `Set` here and as the return value of the method (here and elsewhere in this change set). As a name perhaps just `erroneousCommitLogs`, or similar? No need to encode the type into the variable name.

Thank you for the suggestion! Updated accordingly.

src/main/java/io/debezium/connector/cassandra/KafkaRecordEmitter.java
bingqinzhou: > I'd suggest to _either_ log _or_ rethrow, so not to have the same exception in the logs twice.

Updated as below so that we won't have duplicate logs:
- When we need to throw the error, we throw DebeziumException.
- When we need to catch the error, we log the error message.

src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
bingqinzhou: > Is this set accessed from multiple threads at the same time btw.?

Nice catch! There are multiple threads that can access this set:
- But read access (`set.contains(commitLog-1)`) will always happen after write access (`set.add(commitLog-1)`). 
- Write access (`set.contains(commitLog-1)` and `set.contains(commitLog-2)`) from different threads however can happen at the same time.
I'm not sure if we need to be concerned about thread safe in this case ^ . Do you suggest any safer kind of Set to use please?

gunnarmorling: Assuming there won't be high contention here (i.e. not many concurrent requests at the same time, which I think is not the case, as log files aren't added like every second or so), the simplest is make all methods accessing it `synchronized`. That way, only one thread at a time will access the set.

gunnarmorling:
Thx! Squashed all the commit into one and merged. Taking a step back, instead of passing the "raw" set around, it may make sense to establish some higher-level component (like `ErrorHandler` we do have in core). This then could for instance be marked with the `@ThreadSafe` annotation, and taking care of satisfying that requirement internally. Something to ponder about :)

PR #22, opened 2021-03-17
DBZ-3262

JeremyVigny:
https://issues.redhat.com/browse/DBZ-3262

I tested debezium with cassandra and there is no regression:
- works without SSL
- works with SSL without adding additional properties
- works with SSL with adding additional properties (list of ciphers)

### DIFF
--- a/src/main/java/io/debezium/connector/cassandra/network/SslConfig.java
+++ b/src/main/java/io/debezium/connector/cassandra/network/SslConfig.java
@@ -25,6 +25,8 @@ public class SslConfig {
     public static final String TRUST_MANAGER_ALGORITHM = "trustManager.algorithm";
     public static final String DEFAULT_TRUST_MANAGER_ALGORITHM = "SunX509";
 
+    public static final String CIPHER_SUITES = "cipherSuites";
+
     private Properties configs;
 
     public SslConfig(Properties configs) {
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Very nice! I think it looks good, I'll defer to @bingqinzhou and @ahmedjami for the details. In the meantime, could you please:

* Rework the commit so it doesn't have the "octocat" icon next to it (i.e. only authoring it using a single email address that's linked to your GH account)
* Send a PR for updating the connector docs (https://github.com/debezium/debezium/blob/master/documentation/modules/ROOT/pages/connectors/cassandra.adoc) to add the new attribute at least (it seems the other SSL attributes are missing there, too; purely optional, it'd be lovely if we could have added them all (I know, I'm greedy ;))

ahmedjami:
New Jira Issue to update connector docs => https://issues.redhat.com/browse/DBZ-3312

And I had already checked the changes here with @JeremyVigny, LGTM ;)

bingqinzhou:
LGTM, thanks.

gunnarmorling:
> New Jira Issue to update connector docs => https://issues.redhat.com/browse/DBZ-3312

Thanks! Let's still add the docs for new option right now, so to not further increase the "docs debt".

> And I had already checked the changes here with @JeremyVigny, LGTM ;)

Excellent; so then only the commit authorship needs fixing as discussed above, and we can merge it. Thanks all!

JeremyVigny:
Thanks for your remark @gunnarmorling . I think it's good, I cleaned up the commits

gunnarmorling:
Excellent, applied. Thanks!

PR #24, opened 2021-03-31
[DBZ-336: Do not mark offsets for commit log files with error.

bingqinzhou:
Do not mark offsets for commit log files with error and need to be re-processed.


bingqinzhou:
@gunnarmorling Merging this PR as well since it's just some nit change :)

gunnarmorling:
Hey @bingqinzhou, just seeing this, sorry for the super-long delay. Ideally, there isn't any merge commit, for the sake of a linear, and thus easy to follow history. For that, just use "Rebase and Merge" instead of "Merge" in the GH UI.

PR #25, opened 2021-04-08
[DBZ-3378]: Add the number of change event queues and queue processors.

bingqinzhou:
Enable Cassandra Connector to have multiple change event queues and queue processors for parallel processing.

### DIFF
--- a/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
@@ -240,6 +240,13 @@ public class CassandraConnectorConfig extends CommonConnectorConfig {
             .withDescription(
                     "The amount of time the CommitLogPostProcessor should wait to re-fetch all commitLog files in relocation dir, given in milliseconds. Defaults to 10 seconds (10,000 ms).");
 
+    public static final int DEFAULT_NUM_OF_CHANGE_EVENT_QUEUES = 1;
+    public static final Field NUM_OF_CHANGE_EVENT_QUEUES = Field.create("num.of.change.event.queues")
+            .withType(Type.INT)
+            .withDefault(DEFAULT_NUM_OF_CHANGE_EVENT_QUEUES)
+            .withDescription(
+                    "The number of change event queues and queue processors.");
+
     /**
      * A comma-separated list of fully-qualified names of fields that should be excluded from change event message values.
      * Fully-qualified names for fields are in the form {@code <keyspace_name>.<field_name>.<nested_field_name>}.
@@ -443,6 +450,10 @@ public class CassandraConnectorConfig extends CommonConnectorConfig {
     public Duration commitLogRelocationDirPollInterval() {
         int ms = this.getConfig().getInteger(COMMIT_LOG_RELOCATION_DIR_POLL_INTERVAL_MS);
         return Duration.ofMillis(ms);
+    }
+
+    public int numOfChangeEventQueues() {
+        return this.getConfig().getInteger(NUM_OF_CHANGE_EVENT_QUEUES);
     }
 
     public List<String> fieldExcludeList() {
--- a/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
+++ b/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
@@ -5,12 +5,15 @@
  */
 package io.debezium.connector.cassandra;
 
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.config.Schema;
+import org.apache.kafka.clients.producer.KafkaProducer;
 
 import io.debezium.connector.base.ChangeEventQueue;
 import io.debezium.connector.cassandra.exceptions.CassandraConnectorTaskException;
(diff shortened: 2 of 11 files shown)


gunnarmorling:
Hey, so what are the implications in terms of ordering by using multiple queues? Isn't there a risk for lowering order guarantees here? And did you have a chance to do some testing and evaluating what is gained by this change?

bingqinzhou:
> Hey, so what are the implications in terms of ordering by using multiple queues? Isn't there a risk for lowering order guarantees here? And did you have a chance to do some testing and evaluating what is gained by this change?

The order is guaranteed for change events of the same table or commit log file. I'm using the following logic to make sure that change events from the same table or commit log file will be put into the same queue, and processed by the same queue processor. Does this seem to be a reasonable approach please?

```
queues.get(Math.abs(tableName.hashCode()%queues.size())) :: enqueue
queues.get(Math.abs(commitlogFileName.hashCode()%queues.size())) :: enqueue
```
The ticket for scaling up change events processing was originally created by @joyg. I summarized and replicate the content in  DBZ-3378. A commit log file will stay in cdc_raw directory until all change events associated with this file get processed and published. The gain for this change is that with more threads to process change events, I think the average time of a commit log file staying in cdc_raw directory will be shorten, thus decrease the risk of commit log files piling up in cdc_raw directory, which will potentially cause write suspension in Cassandra DB.

bingqinzhou:
Hi @gunnarmorling, I'm not sure if my explanation above ^ makes sense to you. If you still have concern/question about this change, please let me know. I'd like to provide more context or have more discussion about this, thanks.

gunnarmorling:
> The order is guaranteed for change events of the same table or commit log file.

But that's quite a substantial difference, no? If we can guarantee order per table, we're fine, but order per commit log file seems not enough. After all, you might have changes pertaining to one table in multiple commit log files, so if you process them concurrently, you may end up with events out of order for a specific record. Or am I missing anything?

ahmedjami:
Even we can guarantee order at Cassandra node, it doesn't the case at cluster/datacenter level since the connector should be deployed at least in one datacenter that groups multiple Cassandra replicas, so an individual write is replicated X times due to replication factor.
Users should implement a logic at "kafka level" when they consume messages that put in order the events per table based on Cassandra record writeTime for an X time window.

bingqinzhou:
> > The order is guaranteed for change events of the same table or commit log file.
> 
> But that's quite a substantial difference, no? If we can guarantee order per table, we're fine, but order per commit log file seems not enough. After all, you might have changes pertaining to one table in multiple commit log files, so if you process them concurrently, you may end up with events out of order for a specific record. Or am I missing anything?

I don't think we need to pertain this order, it's not pertained without this PR either. This limitation is documented in https://github.com/debezium/debezium/blob/master/documentation/modules/ROOT/pages/connectors/cassandra.adoc.

```
- Commit logs on an individual Cassandra node do not reflect all writes to the cluster, they only reflect writes stored on that node. This is why it is necesssary to monitor changes on all nodes in a Cassandra cluster. However, due to replication factor, this also implies it is necessary for downstream consumers of these events to handle deduplication.

- Writes to an individual Cassandra node are logged as they arrive. However, these events may arrive out-of-order from which they are issued. Downstream consumers of these events must understand and implement logic similar to Cassandra’s read path to get the correct output.
```

However, we need to guarantee the order of change event within `a table` or `a commit log file` because we need to determine when the last change event of `a table` or `a commit log` file is published, so that we can mark the table or commit log file as done correctly.

gunnarmorling:
Hey, sorry, I completely missed that this PR already got merged. I'm still having doubts here, tbh.

> we need to guarantee the order of change events within a table

How is this ensured with this change? If different commit logs can be processed concurrently, I don't understand how ordering of events within a table would be ensured. Can you elaborate?

Also, are we actually sure that this change is beneficial in terms of throughput? I.e. have you done some load test and measurement which shows that there is an improvement?

ahmedjami:
Hey @gunnarmorling 

In our side, this will not cause any issue. As mentioned above, when we have started using Dbezium we knew that the order will never be guaranteed across the cluster so we have implemented the necessary logic on the downstream side.

I think that what @bingqinzhou would say is that the order will be guaranteed within the same commitLog for the same table. Otherwise, it will not be possible to guarantee this for two differents commit files.

It could be mentioned in the doc, so users can specify the desired num of the queue processors in balance with the order through multiple commit Logs?

gunnarmorling:
Hey Ahmed, thanks for your reply, I begin to understand the reasoning.
Could you send the same reply to my question on the mailing list (I should
have started there to begin with, sorry for the confusion), there's one
follow-up question I'd like to ask, but it's probably better over there
than here on the closed PR. Thx!

>

ahmedjami:
I will send a new PR to update the doc with this new parameter.

gunnarmorling:
Thank you so much! In general, doc changes should always be done as part of
the actual code change PR (i.e. docs should be part of the acceptance
criteria before merging a PR).

>

bingqinzhou:
Thanks @ahmedjami! I'll put up a PR within this sprint to update the doc to include a few recent changes. 
@gunnarmorling Sorry for not updating docs in time! I'll keep this in mind to update doc with code changes going forward.

gunnarmorling:
No worries, @bingqinzhou! Probably best to sync with @ahmedjami as he seems to suggest he may be working on the docs here already, so to avoid any duplicate efforts. Thanks all!

bingqinzhou:
> No worries, @bingqinzhou! Probably best to sync with @ahmedjami as he seems to suggest he may be working on the docs here already, so to avoid any duplicate efforts. Thanks all!

Hi @ahmedjami, please let me know if you already started working on adding the doc for this change :) Otherwise, I'll include this in a doc update PR next week. Thank you!

ahmedjami:
Hi  @bingqinzhou  Not yet, I just created the Jira :) So, yes you can include it in a new doc update PR as the process suggests it.
Thanks!

PR #27, opened 2021-04-20
DBZ-2870 Enforce Java 8

jpechane:
https://issues.redhat.com/browse/DBZ-2870


gunnarmorling:
@jpechane, that build failure is interesting, it reveals that this job is building to much of the main repo. It should build only _debezium-core_ and in turn what this one depends on.

gunnarmorling:
Ha, you beat me to it, just was about to push the very same change but it got rejected :)

jpechane:
@gunnarmorling ou know, someimes I am faster than light ;-)

gunnarmorling:
Admit, you had shelved that change before and only waited with pushing it, so to aggravate me ;)

PR #28, opened 2021-04-29
DBZ-3449 Adding 'quick' profile

gunnarmorling:
https://issues.redhat.com/browse/DBZ-3449

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -22,26 +22,26 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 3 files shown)


gunnarmorling:
@jpechane, there still seems to be some enforcer version issue, as revealed by this (unrelated) PR. Building core with JDK 8 won't work, the GH Action would have to set up two JDKs and use JDK 11 for the main repo and JDK 8 for the Cassandra connector.

jpechane:
@gunnarmorling GH Workflow updated and PR applied, thanks!

PR #31, opened 2021-05-14
DBZ-3523 - Fix exception on not found table

smiklosovic:


bingqinzhou:
From my understanding this doesn't really fix the error about `java.lang.IllegalArgumentException: Unknown keyspace/cf pair (test_keyspace.table2)`, but just return at the place where the error used to be thrown. Since there's already a try catch block which wraps the whole block, an extra try catch shouldn't be needed here. I manually tested it locally and can't reproduce this error. It seems to be only occurring during the test cases in CI build. 

I guess we don't need to worry about this error for now since it only happens during CI build, but I'm still kind of concerned about manually updating the keyspace/table metadata in schema instance. I'm wondering if it would be better to just reload schema.instance by `Schema.instance.loadFromDisk(false)` whenever there's a keyspace/table related change in Cassandra. @smiklosovic Any concerns or opinions about this alternative approach please?

smiklosovic:
Hi @bingqinzhou ,

I put it into a separate try-catch on purpose - in order to distinguish this specific case from the "generic" catch block at the end of the method, otherwise I would mix it together and it would not be clear what exactly has happened and where.

The approach you suggest was explained in my docs - it is not always enough to just load it. There might be cases where you `ALTER` a table with `cdc = true`, Cassandra takes it into account, but it is _not_ persisted on disk yet so you reloading schema does effectively nothing because this change is not _always_ reflected. It is not deterministic, sometimes it manages to persist it before it is reloaded, sometimes not ... From my point of view the only guaranteed way to reflect this change is to have that listener and react on it like we do.

I am thinking about this more deeply and I wonder how that error might happen - it seems like it wants to remove a table which was not added. This might happen either removal has happened before it was added or addition of that table has not succeed for some reason. I wonder what has happened here.

bingqinzhou:
> Hi @bingqinzhou ,
> 
> I put it into a separate try-catch on purpose - in order to distinguish this specific case from the "generic" catch block at the end of the method, otherwise I would mix it together and it would not be clear what exactly has happened and where.
> 
> The approach you suggest was explained in my docs - it is not always enough to just load it. There might be cases where you `ALTER` a table with `cdc = true`, Cassandra takes it into account, but it is _not_ persisted on disk yet so you reloading schema does effectively nothing because this change is not _always_ reflected. It is not deterministic, sometimes it manages to persist it before it is reloaded, sometimes not ... From my point of view the only guaranteed way to reflect this change is to have that listener and react on it like we do.
> 
> I am thinking about this more deeply and I wonder how that error might happen - it seems like it wants to remove a table which was not added. This might happen either removal has happened before it was added or addition of that table has not succeeded for some reason. I wonder what has happened here. Maybe adding and removal, because of the nature of the tests, was done so quickly / in such a fast manner, that removal was actually really processed before addition but it woud not happen in "real world scenarios".
> 
> Please keep in mind that while this is not so much intuitive, the primary reason it is done like this is that if we do not reflect this on a Cassandra's "schema layer" (initialisation of code paths deserialisation deals with), the subsequent deserialisation logic might fail as cdc flag is not set to true yet or schema to deserialise for does not exist and so on ...
> 
> The ultimate solution would be to code up an agent as it would live in one JVM and it would be all transparently visible and you would not have to actually refresh anything. There is `org.apache.cassandra.schema.SchemaChangeListener` one might implement and register into `Schema.instance.registerListener(notifier)` so you would be notified on every change and there you might check if cdc was set to true and act on it by putting (Kafka) schema into our structures and so on. Hence we would abandon this approach. But I do not think that is feasible without some heavy refactoring and even after that we might realise that _dependency-wise_ Debezium just can not live with the class path of Cassandra together and this integration is not possible as such. The other implication by doing this is that we would suddently couple the lifecycle of Debezium to Cassandra and if we wanted to, for example, upgrade the connector, we would have to take that Cassandra node down and so on.

Hi @smiklosovic, thank you for the context and explanation above! 

I'm reading the source codes in `org.apache.cassandra.config.Schema` and wondering if we could just use some public functions in this class directly instead of manually writing codes to update keyspace/table metadata. For example, could we do something like blow:

```
schemaChangeListener = new NoOpSchemaChangeListener() {

@Override            
public void onKeyspaceAdded () {
   *Schema.instance.addKeyspace();*
}

@Override            
public void onKeyspaceChanged() {
  *Schema.instance.updateKeyspace()*
}

@Override            
public void onKeyspaceRemoved() {
  *Schema.instance.dropKeyspace()*
}

@Override            
public void onTableAdded() {
   *Schema.instance.addTable()*
}

@Override            
public void onTableRemoved() {
  *Schema.instance.dropTable()*
}

@Override            
public void onTableChanged() {
  *Schema.instance.updateTable()*
}
```

bingqinzhou:
@smiklosovic After a double thought, I have a question about the approach in PR#26 based on the following situation:

When we run CQL to change the DDL of a table on one node, how long will it take for this DDL change to be propagated to other nodes? If DDL update sync up among Cassandra nodes actually takes a while, say more than 10s, which is the period of retrieving schema changes with Cassandra Client in the original approach, does that mean the delay of DDL detection will be even higher than the original approach on nodes where DDL change is not directly executed on?

smiklosovic:
Hi,

so to your first problem, "Create table_1 again: CREATE TABLE table_1, and the following error will be thrown:"

This is actually easy to debug why it is happening, if you follow the stack trace, you see that it will propagate to this code:

https://github.com/apache/cassandra/blob/181a4969290f1c756089b2993a638fe403bc1314/src/java/org/apache/cassandra/db/ColumnFamilyStore.java#L471-L516

There you see that it attempts to register that mbean for the second time and it throws on line 490. The solution to this does exist but it is not so straighforward. Firstly, I would have to call that method with `registerBookkeeping` set to false so that means I would have to use some reflection because it is used here:

https://github.com/apache/cassandra/blob/181a4969290f1c756089b2993a638fe403bc1314/src/java/org/apache/cassandra/db/Keyspace.java#L417

but that `columnFamilyStores` is private and getter returns unmodifiable collection. 

The second throw is the result of the adding failed so my hypothesis with failed adding was actually correct, thanks for giving me these stacktraces.

"wondering if we could just use some public functions in this class directly instead of manually writing codes" - I tried the same at first but you realise that it is not possible to reuse without node running. It it is invoking methods, for example:

```
    public void addTable(CFMetaData cfm)
    {
        assert getCFMetaData(cfm.ksName, cfm.cfName) == null;

        // Make sure the keyspace is initialized
        // and init the new CF before switching the KSM to the new one
        // to avoid races as in CASSANDRA-10761
        Keyspace.open(cfm.ksName).initCf(cfm, true);
        // Update the keyspaces map with the updated metadata
        update(cfm.ksName, ks -> ks.withSwapped(ks.tables.with(cfm)));
        // Update the table ID <-> table name map (cfIdMap)
        load(cfm);
        MigrationManager.instance.notifyCreateColumnFamily(cfm);
    }
```
Do you see the last statement? That one would cause error in our scenario. So for that reason I am doing it like I extract all stuff from there but without bits which are problematic. Same stuff might apply for all other cases.

DDL propagation is rather delicate topic. This was rewritten lately (my collab with some other dev) here 

https://github.com/apache/cassandra/commit/a960d87e05f01000a758d8e7e5a58be57d13eb33

It depends what version of Cassandra one uses, I think it was changed in 3.11.9. It is quite broad topic (I might explain that to you if you insist in more depth) but to answer your question, it works like each node is periodically checking schema versions of other nodes via EndpointState via Gossip protocol and if it detect that other nodes are on a different schema version, it will pull these schemas from them and if it pulls schemas from all nodes, by definition, it will be on the latest schema version. There is a lot of moving parts as "when that happens" but I think it schedules a task which is executed every minute but I would have to double check.

I do not think that returning to periodic fetching instead of a listener as we have it now would bring any benefit in this regard. Think about it, if you periodically check if a node has so and so schema (as it was done before), it means that such schema - reflecting some (cdc) change - would have to be already present in the respective node you are fetching it from - and this schema version is propagated to it _eventually_. So it is irrelevant if you are actively fetching it every n seconds or if a listener is trigger once that happens because you still need to wait for that schema propagation in the node from other nodes anyway.

bingqinzhou:
@smiklosovic Thank you for the context and explanation, these make a lot of sense to me! As for the error `java.lang.IllegalArgumentException: Unknown keyspace/cf pair (test_keyspace.table2)`, since it's tricky to fix and the situation only happens in unit tests, I don't think we need to worry about it atm :)

gunnarmorling:
@smiklosovic, @bingqinzhou, so is this one good to merge? I'm still not up to speed on all these changes related to schema changes, so relying on your expertise here really. Thanks a lot!

smiklosovic:
Hi @bingqinzhou ,

I have pushed some changes, it might be achieved by doing this in "onTableRemoved", is fairly simple:

```
final Method unregisterMBeanMethod = ColumnFamilyStore.class.getDeclaredMethod("unregisterMBean");
unregisterMBeanMethod.setAccessible(true);
unregisterMBeanMethod.invoke(cfs);
```

I think this solves the create - drop - create issue well!

Regards

src/main/java/io/debezium/connector/cassandra/SchemaProcessor.java
```diff
@@ -139,34 +141,45 @@ public void onTableRemoved(final TableMetadata table) {
...
                     // reinitialize the keyspace.
-                    final CFMetaData cfm = oldKsm.tables.get(tableName).get();
-                    final org.apache.cassandra.schema.KeyspaceMetadata newKsm = oldKsm.withSwapped(oldKsm.tables.without(tableName));
+                    final Optional<CFMetaData> cfm = oldKsm.tables.get(tableName);
 
-                    Schema.instance.unload(cfm);
-                    Schema.instance.setKeyspaceMetadata(newKsm);
+                    final Method unregisterMBeanMethod = ColumnFamilyStore.class.getDeclaredMethod("unregisterMBean");
+                    unregisterMBeanMethod.setAccessible(true);
+                    unregisterMBeanMethod.invoke(cfs);
```
gunnarmorling, line 169: So that'll be quite troublesome when moving to newer Java versions (once supported by Cassandra). What's the reasoning for doing this reflectively? A comment in the code would be warranted at the very least, and we'll have to look for alternatives going forward.

src/main/java/io/debezium/connector/cassandra/SchemaProcessor.java
```diff
@@ -139,34 +141,45 @@ public void onTableRemoved(final TableMetadata table) {
...
-                    final ColumnFamilyStore cfs = Keyspace.openWithoutSSTables(ksName).getColumnFamilyStore(tableName);
+                    ColumnFamilyStore cfs;
 
-                    if (cfs == null) {
-                        LOGGER.warn("ColumnFamilyStore for {}.{} is not found!", table.getKeyspace(), table.getName());
+                    try {
+                        cfs = Keyspace.openWithoutSSTables(ksName).getColumnFamilyStore(tableName);
+                    }
+                    catch (final Exception ex) {
+                        LOGGER.warn("ColumnFamilyStore for {}.{} is not found!", ksName, table.getName());
```
gunnarmorling, line 157: In which situation would this happen? A comment in the code would be great.

gunnarmorling (commented):
I realize this has been merged already, but two questions inline. Thanks!

src/main/java/io/debezium/connector/cassandra/SchemaProcessor.java
smiklosovic: Why would moving to newer Java cause a problem here? Btw afaik Cassandra supports 8 and 11 now only. (and 11 only in case of Cassandra 4 this connector in incompatible anyway yet).

gunnarmorling: In newer version, reflection is more and more frowned upon. Java 16 by default disallows reflection on JDK internal types, the same when trying to do this with modules (whenever that'd be the case here). So it's not an immediate problem, but the more we can avoid this sort of reflective access, the better. So why is it this method needs to be invoked that way?

src/main/java/io/debezium/connector/cassandra/SchemaProcessor.java
smiklosovic: Right now, after we fixed it, we should be pretty safe to never encounter this again. This is more to be just sure to log it but I cant see how it might happen again.

src/main/java/io/debezium/connector/cassandra/SchemaProcessor.java
smiklosovic: That method is package protected so we would have to accommodate the code to `org.apache.cassandra.db`. I mean, be might indeed do that if you dont mind shuffling with the code here and there.

The way how it would look like is like this:

https://github.com/debezium/debezium-connector-cassandra/pull/33

gunnarmorling: Replied over there; thx.

src/main/java/io/debezium/connector/cassandra/SchemaProcessor.java
gunnarmorling: Ok, thx.

gunnarmorling:
@bingqinzhou, thanks for merging this! One thing only: let's use rebases over merge conflicts generally, for the sake of a linear (and thus easier to grasp) history.

PR #33, opened 2021-05-25
moved nooplistener to apache package

smiklosovic:


src/main/java/org/apache/cassandra/db/NoOpSchemaChangeListener.java
```diff
@@ -3,7 +3,7 @@
  *
  * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
  */
-package io.debezium.connector.cassandra;
+package org.apache.cassandra.db;
```
gunnarmorling, line 6: Ok, let's definitely not do that :) The split package will haunt us eventually, and it's generally not a good idea to inject code into packages not owned by us.

So taking a step back, _why_ is it that we need to unregister that MBean in the first place? If there's no public method for doing so, this makes me think it's not something we actually should do. What about invoking `ColumnFamilyStore.invalidate()`, which also handles the MBean business?

smiklosovic: adding a table also registers mbean but removing a table does not unregister it, so then the next adding will fail because it will try to register that mbean for the second time which fails

the only reasonable way how to do that, as what I found, is to actually unregister it - that is the proper way, but that method is package protected :)

Invocation of invalidate fails too.

bingqinzhou:
@smiklosovic Trying to understand the context here, why is this change needed after the fix in https://github.com/debezium/debezium-connector-cassandra/pull/31?

src/main/java/org/apache/cassandra/db/NoOpSchemaChangeListener.java
gunnarmorling: > Invocation of invalidate fails too.

Do you have any more details why it fails?

gunnarmorling:
> Trying to understand the context here

@bingqinzhou, the question is whether there's ways to avoid that reflective invocation of the method for unregistering that MBean.

smiklosovic:
@bingqinzhou we just discuss things ... Gunnar does not like reflection but without it we would have to move it to another package because `unregister` is package protected, this just models how it would be like for him to see and he eventually rejected that idea anyway.

PR #56, opened 2022-05-18
DBZ-5143 initialize database properly before tests

vjuranek:
https://issues.redhat.com/browse/DBZ-5143


smiklosovic:
@vjuranek you do not need to do this. This is what I was trying to do, that, interestingly, for that test to pass, you do not need to start Cassandra at all :) I fixed it differently, it is included in the patch for Cassandra 3.

vjuranek:
Makes sense @smiklosovic, closing this one.

PR #57, opened 2022-05-18
DBZ-4401 apply changes to Cassandra 3 implementation

smiklosovic:


smiklosovic:
@vjuranek this is the patch for Cassandra 3. It will also fix the build issue currently in main branch. Not sure what that comes from ... I ve just rewrote it to somethign simpler.

cassandra-3/src/main/java/io/debezium/connector/cassandra/Cassandra3CommitLogReadHandlerImpl.java
```diff
@@ -109,7 +120,9 @@ enum PartitionType {
          */
         COUNTER;
 
-        static final Set<PartitionType> supportedPartitionTypes = new HashSet<>(Arrays.asList(PARTITION_KEY_ROW_DELETION, ROW_LEVEL_MODIFICATION));
```
smiklosovic, line 112: @vjuranek in the future we way think about extracting this stuff to some common helper class, same with supportedPartitionTypes and related things. I just left it here as is now, we might realise that some operations are doable in Cassandra 4 only (or vice versa) so having these variables "global" is not a good idea _yet_.

core/src/test/java/io/debezium/connector/cassandra/TestUtils.java
```diff
@@ -142,9 +141,8 @@ public static String keyspaceTable(String tableName) {
      * Generate a commit log file with current timestamp in memory
      */
     public static File generateCommitLogFile() {
-        File cdcDir = new File(DatabaseDescriptor.getCDCLogLocation());
         long ts = System.currentTimeMillis();
-        return Paths.get(cdcDir.getAbsolutePath(), "CommitLog-6-" + ts + ".log").toFile();
+        return Paths.get("/tmp", "CommitLog-6-" + ts + ".log").toFile();
```
vjuranek, line 145: `/tmp` os OS specific, you can use something like `Files.createTempDirectory("cassandra_test_").toString()` to make it OS agnostic (also is more safe as it should create different dir for each test).

vjuranek (approved):
One minor thing, otherwise LGTM.

core/src/test/java/io/debezium/connector/cassandra/TestUtils.java
smiklosovic: it really does not do anything, it does not matter. it will not place any files there and so on ... it is irrelevant how it is called.

vjuranek:
@ahmedjami would you like to test also this change or you consider testing only Cassandra 4 sufficient enough?

core/src/test/java/io/debezium/connector/cassandra/TestUtils.java
vjuranek: well, but it's a generic test utility which may be used in the future in other places and in a wrong way. We either should generate existing path (which would work on every OS) or return just a constant or random string and name the method accordingly. IMHO the best at this point would be to remove this method completely and use some string (e.g. `non-existing-log-file-path`) directly in the test.

smiklosovic: @vjuranek  fixed and pushed, lets see.

vjuranek (approved):

vjuranek:
Thanks @smiklosovic , LGTM now. @ahmedjami please let us know once you finish testing, we need to merge this to unbreak the build. If it's going to take some longer time, please let us know as well, so we can fix the build in separate PR. Thanks!

ahmedjami:
Yes, it's in progress and it shouldn't it shouldn't take long :)

ahmedjami:
Hey @vjuranek 
I have just finished testing with Cassandra-3 and all is working as expected :)
Greetings,

vjuranek:
Merged. Again, thanks @smiklosovic for PR and @ahmedjami for testing!

jpechane:
@vjuranek Will you backport the PR against 1.9?

vjuranek:
@jpechane backported

PR #64, opened 2022-07-25
DBZ-5380 - fixed schema changed listener for keyspaces and tables

smiklosovic:


Naros:
Hi @smiklosovic, is this meant to target 1.9 or should this be targetting `main` at this point?

smiklosovic:
Hi @Naros , I believe this should be included from 1.9 (included) up to main.

smiklosovic:
@Naros I have changed this PR in such a way that this branch is for `main`, you just need to apply to to lower branches.

Naros:
> @Naros I have changed this PR in such a way that this branch is for `main`, you just need to apply it to lower branches.

Awesome, thanks @smiklosovic

core/src/test/java/io/debezium/connector/cassandra/TestUtils.java
```diff
@@ -89,19 +99,28 @@ public static List<String> getTables(CqlSession session) {
...
+
+    public static void deleteTestKeyspaceTables(String keyspaceName) throws Exception {
         try (CqlSession session = CqlSession.builder().build()) {
-            for (String table : getTables(session)) {
-                session.execute(SimpleStatement.newInstance(String.format("DROP TABLE IF EXISTS %s.%s", TEST_KEYSPACE_NAME, table)));
+            for (String table : getTables(keyspaceName, session)) {
+                session.execute(SimpleStatement.newInstance(String.format("DROP TABLE IF EXISTS %s.%s", keyspaceName, table)));
             }
         }
+        Thread.sleep(5000);
```
jpechane, line 123: Why is the sleep needed here? Could it be replaced by a condition to check?

smiklosovic: it is there for schema listener in debezium to have a chance to react on the schema changes in Cassandra. 5s on keyspace creation seems to be like a safe value. I do not think waiting by a "condition" is appropriate as I do not know what that condition would actually wait for in this context.

jpechane: ok, I understand what's missing/is difference from what I expected. In case of RDBMS we have `DatabaseHistoryMetrics` that is set of JMX metrics allowing user to monitor schema changes. Maybe Cassandra should provide a similar instrument. In that case it would be also pretty easy to check the metrics and wait for change to be reacted on.

smiklosovic: probably, we might do this on top of the current pr after it is merged, I do not think the change you are suggesting is related to this PR as such.

jpechane: Aggred, there is follow-up Jira for that - https://issues.redhat.com/browse/DBZ-5523

jpechane:
@smiklosovic Applied, thanks!

PR #62, opened 2022-07-20
DBZ-5410: Ability to read commit log files incrementally in Cassandra4

nitinitt:
…4 and not when its marked Complete

PR #69, opened 2022-08-19
DBZ-5410 Added ability to process Cassandra 4 Commit log files incrementally

nitinitt:
I had a merge commit in my previous PR: https://github.com/debezium/debezium-connector-cassandra/pull/62 and am not able to remove that without altering the git history of "main" branch. Please accept this PR as part of: 
https://issues.redhat.com/browse/DBZ-5410


nitinitt:
@jpechane @smiklosovic  re-tagging you in this PR.

smiklosovic:
I ll do a review next week.

cassandra-4/src/main/java/io/debezium/connector/cassandra/AbstractCassandra4CommitLogParser.java
```diff
@@ -0,0 +1,116 @@
...
+    protected final CommitLogProcessorMetrics metrics;
+    private final Cassandra4CommitLogReadHandlerImpl commitLogReadHandler;
+
+    private final CommitLogTransfer commitLogTransfer;
+    private final Set<String> erroneousCommitLogs;
+    protected boolean completePrematurely = false;
+    protected Cassandra4CommitLogProcessor.LogicalCommitLog commitLog;
+    protected int pollingInterval;
+
+    public AbstractCassandra4CommitLogParser(Cassandra4CommitLogProcessor.LogicalCommitLog commitLog, final List<ChangeEventQueue<Event>> queues,
```
smiklosovic, line 35: Could you please move `LogicalCommitLog` out of `Cassandra4CommitLogProcessor` so it is standalone class in the same package? I find it weird that parser depends on some random stuff in log processor. Same situation with `ProcessingResult`.

cassandra-4/src/main/java/io/debezium/connector/cassandra/AbstractCassandra4CommitLogParser.java
```diff
@@ -0,0 +1,116 @@
...
+    public void complete() {
+        completePrematurely = true;
+    }
+
+    public abstract Cassandra4CommitLogProcessor.ProcessingResult parse();
+
+    public Cassandra4CommitLogProcessor.ProcessingResult process() {
+        if (!commitLog.exists()) {
+            LOGGER.warn("Commit log " + commitLog + " does not exist!");
+            return new Cassandra4CommitLogProcessor.ProcessingResult(commitLog, Cassandra4CommitLogProcessor.ProcessingResult.Result.DOES_NOT_EXIST);
```
smiklosovic, line 66: `DOES_NOT_EXIST` might be imported.

cassandra-4/src/main/java/io/debezium/connector/cassandra/AbstractCassandra4CommitLogParser.java
smiklosovic: ProcessingResult might be renamed to `CommitLogProcessingResult`.

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogBatchParser.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public Cassandra4CommitLogProcessor.ProcessingResult parse() {
+        try {
+            parseIndexFile(commitLog);
+
+            while (!commitLog.completed) {
+                if (completePrematurely) {
+                    LOGGER.info("{} completed prematurely", commitLog);
```
smiklosovic, line 35: might be `{} processing completed prematurely` or `Processing of {} completed prematurely`.

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
```diff
@@ -0,0 +1,79 @@
...
+                                             CommitLogProcessorMetrics metrics, CassandraConnectorContext context) {
+        super(commitLog, queues, metrics, context);
+    }
+
+    @Override
+    public Cassandra4CommitLogProcessor.ProcessingResult parse() {
+        try {
+            parseIndexFile(commitLog);
+            while (!commitLog.completed) {
+                LOGGER.info("Polling for completeness of idx file for: {}", commitLog.toString());
```
smiklosovic, line 32: `toString` is not necessary.

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
```diff
@@ -0,0 +1,79 @@
...
+                if (completePrematurely) {
+                    LOGGER.info("{} completed prematurely", commitLog.toString());
+                    return new Cassandra4CommitLogProcessor.ProcessingResult(commitLog, Cassandra4CommitLogProcessor.ProcessingResult.Result.COMPLETED_PREMATURELY);
+                }
+
+                CommitLogPosition commitLogPosition = null;
+                if (offset == null) {
+                    LOGGER.info("Start to read the partial file : {}", commitLog.toString());
```
smiklosovic, line 40: `commitLog.toString` is not necessary

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
```diff
@@ -0,0 +1,79 @@
...
+                    commitLogPosition = new CommitLogPosition(commitLog.commitLogSegmentId, 0);
+                }
+                else if (offset < commitLog.offsetOfEndOfLastWrittenCDCMutation) {
+                    LOGGER.info("Resume to read the partial file : {}", commitLog.toString());
```
smiklosovic, line 44: `commitLog.toString` is not necessary

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
```diff
@@ -0,0 +1,79 @@
...
+                    commitLogPosition = new CommitLogPosition(commitLog.commitLogSegmentId, offset);
+                }
+                else {
+                    LOGGER.info("No movement in offset in IDX file: {}", commitLog.toString());
```
smiklosovic, line 48: `commitLog.toString` is not necessary

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
```diff
@@ -0,0 +1,79 @@
...
+                    offset = commitLog.offsetOfEndOfLastWrittenCDCMutation;
+                    metrics.setCommitLogPosition(commitLogPosition.position);
+                }
+
+                LOGGER.info("Sleep for idx file to be complete");
+                Thread.sleep(pollingInterval);
+                parseIndexFile(commitLog);
+            }
+
+            LOGGER.info("IDX file is completed for: {}", commitLog.toString());
```
smiklosovic, line 62: Once there is `IDX` other time `idx`. We should be consistent in it. Also, `commitLog.toString` here is not necessary.

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
```diff
@@ -0,0 +1,79 @@
...
+            }
+            else {
+                commitLogPosition = new CommitLogPosition(commitLog.commitLogSegmentId, 0);
+            }
+            metrics.setCommitLogPosition(commitLogPosition.position);
+            processCommitLog(commitLog, commitLogPosition);
+            return new Cassandra4CommitLogProcessor.ProcessingResult(commitLog);
+        }
+        catch (final Exception ex) {
+            LOGGER.error("Processing of {} errored out", commitLog.toString(), ex);
```
smiklosovic, line 75: also, toString is redundant

core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -150,6 +150,32 @@ public static Optional<SnapshotMode> fromText(String text) {
...
             .withDescription("Determines whether or not the CommitLogProcessor should re-process error commitLogFiles.");
 
+    /**
+     * Only valid for Cassandra 4 and if enabled, commit logs would be read incrementally instead of reading complete log file
+     */
+    public static final boolean DEFAULT_COMMIT_LOG_REAL_TIME_PROCESSING_ENABLED = false;
+    public static final Field COMMIT_LOG_REAL_TIME_PROCESSING_ENABLED = Field.create("commit.log.real.time.processing.enabled")
+            .withType(Type.BOOLEAN)
+            .withDefault(DEFAULT_COMMIT_LOG_REAL_TIME_PROCESSING_ENABLED)
+            .withDescription("Enables the near real time processing of commit logs for Cassandra 4 by reading commit log files incrementally");
```
smiklosovic, line 160: should not there be some hyphens in near real time? Like `near real-time`.

core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -150,6 +150,32 @@ public static Optional<SnapshotMode> fromText(String text) {
...
+    }
+
+    /**
+     * Only valid for Cassandra 4 and defines the polling interval to check for completeness of commit log file
+     */
+    public static final int DEFAULT_COMMIT_LOG_MARKED_COMPLETE_POLL_INTERVAL_IN_MS = 10_000;
+    public static final Field COMMIT_LOG_MARKED_COMPLETE_POLL_INTERVAL_IN_MS = Field.create("commit.log.marked.complete.poll.interval.ms")
+            .withType(Type.INT)
+            .withDefault(DEFAULT_COMMIT_LOG_MARKED_COMPLETE_POLL_INTERVAL_IN_MS)
+            .withDescription("Defines the polling interval to check for Commit Log file marked complete in Cassandra 4");
```
smiklosovic, line 173: I think it should be `CommitLog`, like these words should be together.

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public Cassandra4CommitLogProcessor.ProcessingResult parse() {
+        try {
+            parseIndexFile(commitLog);
+            while (!commitLog.completed) {
+                LOGGER.info("Polling for completeness of idx file for: {}", commitLog.toString());
+                if (completePrematurely) {
+                    LOGGER.info("{} completed prematurely", commitLog.toString());
+                    return new Cassandra4CommitLogProcessor.ProcessingResult(commitLog, Cassandra4CommitLogProcessor.ProcessingResult.Result.COMPLETED_PREMATURELY);
```
smiklosovic, line 35: might be imported

cassandra-4/src/main/java/io/debezium/connector/cassandra/AbstractCassandra4CommitLogParser.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    protected void processCommitLog(Cassandra4CommitLogProcessor.LogicalCommitLog logicalCommitLog, CommitLogPosition position) {
+        try {
+            LOGGER.debug("starting to read commit log segments {} on position {}", logicalCommitLog, position);
+            commitLogReader.readCommitLogSegment(commitLogReadHandler, logicalCommitLog.log, position, false);
+            LOGGER.debug("finished reading commit log segments {} on position {}", logicalCommitLog, position);
+        }
+        catch (Exception e) {
+            if (commitLogTransfer.getClass().getName().equals(CassandraConnectorConfig.DEFAULT_COMMIT_LOG_TRANSFER_CLASS)) {
```
smiklosovic, line 97: DEFAULT_COMMIT_LOG_TRANSFER_CLASS might be imported

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
smiklosovic: Also, can you please remove redundant space before the colon? Like `... partial file: {}`.

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogRealTimeParser.java
smiklosovic: plus same space issue before colon.

cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4CommitLogProcessor.java
```diff
@@ -96,23 +94,25 @@ public void stop() {
...
-        Future<ProcessingResult> future = executorService.submit(callable);
-        submittedProcessings.add(new Pair<>(callable, future));
+        if (context.getCassandraConnectorConfig().isCommitLogRealTimeProcessingEnabled()) {
+            parser = new Cassandra4CommitLogRealTimeParser(new LogicalCommitLog(index.toFile()), queues, metrics, this.context);
+        }
+        else {
+            parser = new Cassandra4CommitLogBatchParser(new LogicalCommitLog(index.toFile()), queues, metrics, this.context);
+        }
+
+        Future<ProcessingResult> future = executorService.submit(() -> parser.process());
```
smiklosovic, line 114: `() -> parser.process()` might be replaced with method reference like `.submit(parser::proces)`.

cassandra-4/src/main/java/io/debezium/connector/cassandra/AbstractCassandra4CommitLogParser.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+        catch (InterruptedException e) {
+            throw new CassandraConnectorTaskException(String.format(
+                    "Enqueuing has been interrupted while enqueuing EOF Event for file %s", commitLog.log.getName()), e);
+        }
+    }
+
+    protected void processCommitLog(Cassandra4CommitLogProcessor.LogicalCommitLog logicalCommitLog, CommitLogPosition position) {
+        try {
+            LOGGER.debug("starting to read commit log segments {} on position {}", logicalCommitLog, position);
```
smiklosovic, line 92: Could you upper-case this? I am sorry I left it like that (I think it was me). Same with the next debug log below.

smiklosovic:
The main thing I am missing here is a proper test case. Without this, we can not know this actually works without testing that manually. I think the solid start would be to try to create a simple test case, something like this:

	public class RealtimeTest extends AbstractCommitLogProcessorTest {

	    @Override
	    protected CassandraConnectorContext generateTaskContext() throws Exception {
		Properties properties = TestUtils.generateDefaultConfigMap();
		properties.put(CassandraConnectorConfig.COMMIT_LOG_REAL_TIME_PROCESSING_ENABLED, true);
		return generateTaskContext(Configuration.from(properties));
	    }

	    @Override
	    public void initialiseData() throws Exception {

	    }

	    @Override
	    public void verifyEvents() throws Exception {

	    }
	}

`AbstractCommitLogProcessorTest` is currently written in quite opinionated way and it does not count on the fact we might initialise and verify events differently. I think we might create another abstract class which satisfies real-time testing scenario better with refactored `setup` method and so on. That is where you might get quite creative, I have not explored that more deeply.

My idea of a test case is something like this:

* insert few rows and wait until they are processed (like 10s or what is there, we might also change that config value for the sake of tests being faster / scanning occuring more often).
* verify that rows were read
* insert yet few more rows
* verify only these added rows were added

I think that after some refactoring this should be quite straightforward to do.

nitinitt:
> The main thing I am missing here is a proper test case. Without this, we can not know this actually works without testing that manually. I think the solid start would be to try to create a simple test case, something like this:
> 
> ```
> public class RealtimeTest extends AbstractCommitLogProcessorTest {
> 
>     @Override
>     protected CassandraConnectorContext generateTaskContext() throws Exception {
> 	Properties properties = TestUtils.generateDefaultConfigMap();
> 	properties.put(CassandraConnectorConfig.COMMIT_LOG_REAL_TIME_PROCESSING_ENABLED, true);
> 	return generateTaskContext(Configuration.from(properties));
>     }
> 
>     @Override
>     public void initialiseData() throws Exception {
> 
>     }
> 
>     @Override
>     public void verifyEvents() throws Exception {
> 
>     }
> }
> ```
> 
> `AbstractCommitLogProcessorTest` is currently written in quite opinionated way and it does not count on the fact we might initialise and verify events differently. I think we might create another abstract class which satisfies real-time testing scenario better with refactored `setup` method and so on. That is where you might get quite creative, I have not explored that more deeply.
> 
> My idea of a test case is something like this:
> 
> * insert few rows and wait until they are processed (like 10s or what is there, we might also change that config value for the sake of tests being faster / scanning occuring more often).
> * verify that rows were read
> * insert yet few more rows
> * verify only these added rows were processed
> 
> I think that after some refactoring this should be quite straightforward to do.

Hey @smiklosovic Thanks for the pointers on creating tests, and I totally agree to have tests to have confidence in this new feature. I was able to add a new test for real-time parsing, exactly as you mentioned in the comment above. It was little trickier than I thought particularly due to presence of 2 Cassandra YAML file(one for embedded cassandra/docker and other for running debezium tests). 
Also, have addressed all the comments as well. Hopefully the PR is in a mergeable state now.

cassandra-4/src/test/java/io/debezium/connector/cassandra/CommitLogRealTimeParserTest.java
```diff
@@ -0,0 +1,99 @@
...
+            count += 1;
+        }
+
+        LOGGER.info("Total events received: {}", events.size());
+        Assert.assertEquals("Total number of events received must be " + expectedEventsCount, expectedEventsCount, events.size());
+
+        for (int i = 0; i < expectedEventsCount; i++) {
+            Record record = (Record) events.get(i);
+            Assert.assertEquals("Operation type must be insert", Record.Operation.INSERT, record.getOp());
+            Assert.assertEquals("Inserted key should be " + i, record.getRowData().getPrimary().get(0).value, i + keyInc);
```
smiklosovic, line 78: should not this be `"Inserted key should be" + (i + keyInc)` ?

cassandra-4/src/test/java/io/debezium/connector/cassandra/CommitLogRealTimeParserTest.java
```diff
@@ -0,0 +1,99 @@
...
+            events.addAll(queue.poll());
+            count += 1;
+        }
+
+        LOGGER.info("Total events received: {}", events.size());
+        Assert.assertEquals("Total number of events received must be " + expectedEventsCount, expectedEventsCount, events.size());
+
+        for (int i = 0; i < expectedEventsCount; i++) {
+            Record record = (Record) events.get(i);
+            Assert.assertEquals("Operation type must be insert", Record.Operation.INSERT, record.getOp());
```
smiklosovic, line 77: would be nice to actually log what that operation was if it was not insert, it is nice we know it is not insert, so what is it?

cassandra-4/src/test/java/io/debezium/connector/cassandra/CommitLogRealTimeParserTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Override
+    public void readLogs() throws Exception {
+        // check to make sure there are no records in the queue to begin with
+        assertEquals(queue.totalCapacity(), queue.remainingCapacity());
+        String cdcLoc = DatabaseDescriptor.getCDCLogLocation();
+        LOGGER.info("CDC Location: {}", cdcLoc);
+
+        File[] commitLogs = CommitLogUtil.getIndexes(new File(cdcLoc));
+        Thread.sleep(2000);
```
smiklosovic, line 90: what is this sleep good for? Would you mind to elaborate a bit? Maybe it would be better to use `Awaitility` and wait for some condition to happen? Having explicit sleep times like this tend to be flaky from what I observed in the past.

cassandra-4/src/test/resources/docker/cassandra.yaml
```diff
@@ -222,7 +222,7 @@ cdc_enabled: true
 # separate spindle than the data directories. If not set, the default directory is
 # $CASSANDRA_HOME/data/cdc_raw.
 # cdc_raw_directory: /var/lib/cassandra/cdc_raw
-cdc_raw_directory: /var/lib/cassandra/cdc_raw_directory
+cdc_raw_directory: /var/lib/cassandra/cdc_raw
```
smiklosovic, line 225: just double checking, is this what we want? (in case it was a typo)

cassandra-4/src/test/java/io/debezium/connector/cassandra/CommitLogRealTimeParserTest.java
nitinitt: Glad you asked. Firstly, this sleep is at incorrect place. It should be above L89, i.e. before reading the idx files, and you are right this sleep can cause flakiness in tests and I have replaced it with Awaitility. Cassandra periodically(as defined in cassandra.yaml file, param: commitlog_sync_period_in_ms) flushes idx files and Commit Log files to cdc location. The sleep was to make sure idx file got generated/updated, and now it will be replaced by a condition.

cassandra-4/src/test/resources/docker/cassandra.yaml
nitinitt: Yes, this was intentional.

cassandra-4/src/test/java/io/debezium/connector/cassandra/CommitLogRealTimeParserTest.java
nitinitt: done

cassandra-4/src/test/java/io/debezium/connector/cassandra/CommitLogRealTimeParserTest.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public void readLogs() throws Exception {
+        // check to make sure there are no records in the queue to begin with
+        assertEquals(queue.totalCapacity(), queue.remainingCapacity());
+        String cdcLoc = DatabaseDescriptor.getCDCLogLocation();
+        LOGGER.info("CDC Location: {}", cdcLoc);
+
+        File[] commitLogs = CommitLogUtil.getIndexes(new File(cdcLoc));
```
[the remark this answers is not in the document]
nitinitt: Yes, you are right. Have renamed the variable to commitLogIndexes for better readability

cassandra-4/src/test/java/io/debezium/connector/cassandra/CommitLogRealTimeParserTest.java
```diff
@@ -0,0 +1,99 @@
...
+        verify(3, 10);
+        insertRows(2, 20);
+        verify(2, 20);
+    }
+
+    private void verify(int expectedEventsCount, int keyInc) throws InterruptedException {
+        List<Event> events = queue.poll();
+        int count = 0;
+        int maxRetryCount = 3;
+        while (events.size() != expectedEventsCount && count < maxRetryCount) {
```
jpechane, line 67: Please use `Awaitility` library for this type of code like we do in other connectors.

nitinitt: Thanks, updated the test to use Awaitility.

cassandra-4/src/test/java/io/debezium/connector/cassandra/CommitLogRealTimeParserTest.java
nitinitt: This sounds better, have updated the assert statement. Thanks!

nitinitt:
Hey @jpechane, I have incorporated all the code review feedback. Please let me know if its in a mergeable/acceptable state.

jpechane:
@nitinitt Applied, thanks a lot! Could you please provide a docs PR to the main repo reflecting the change in the connector behaviour? Thanks

nitinitt:
> @nitinitt Applied, thanks a lot! Could you please provide a docs PR to the main repo reflecting the change in the connector behaviour? Thanks

Thanks @jpechane for merging the PR. I will update the docs and create a new PR in Debezium repo. Would this feature be a part of 2.0.0.beta2 release? Would you want this PR to be applied to 1.9 series as well?

jpechane:
@nitinitt Yes, it will be in 2.0.0.Beta2. There is no need for 1.9 as this is a new feature and we usually limit backporting to bugfixes only.

PR #72, opened 2022-09-22
DBZ-5640 literal "${project.version}" in the source record

ggaborg:
DBZ-5640 literal "${project.version}" in the source record instead of the actual version


jpechane:
@ggaborg Applied, thanks

PR #75, opened 2022-10-11
DBZ-2098 Expose Cassandra Connector via Debezium Server

ggaborg:
Hi, this change takes a minimalistic approach to make cassandra connector compatible with debezium server without changing its current functionality. With that in mind, there are few important notes:
- the connector can be run as standalone java application (as today) or inside debezium server as any other connector
- debezium server must be run inside each node running cassandra just like the standalone version
- it is limited to its own implementation of file offset writer (for debezium server the memory offset backing store needs to be defined in the application.properties configuration file even if it not used - see README.md)
- several java options need to be added at startup - see README.md
- it is compatible with all sink types
- example of application.properties for debezium server in README.md
- building the connector with "-P assembly" profile will build the debezium compatible connector plugin instead of the standalone java application (which does not require any profile at build time) 
- building both the connector plugin and the standalone java application at the same time is not possible because of several  dependencies which are defined as provided (by debezium) in case of the connector plugin


jpechane:
@ggaborg Thanks a lot, this PR will be handle after 2.0.0 final release. Thanks for the patience.

ggaborg:
@jpechane Hi, how can we advance with this? Our team has tested a lot, we found some bugs and fixed them in this PR. It looks like it is good to be merged. Thanks!

spicy-sauce:
> @ggaborg Thanks a lot, this PR will be handle after 2.0.0 final release. Thanks for the patience.

Hi @jpechane, what's the timeline here for reviewing and merging this PR? Thanks a lot!

jpechane:
@smiklosovic Hi, could you please take a look at this one? Would it make any significant impact on your operations? The intent is to align with the rest of Debezium codebase and provide full-backward compatibility in behaviour.

smiklosovic:
@jpechane I ll take a look.

README.md
```diff
@@ -35,6 +35,54 @@ Please follow the steps described in the main repository to build the connector:
...
+```
+
+## Sample of `application.properties` for running inside Debezium Server
+
+```
+# Sink
+debezium.sink.type=redis
+debezium.sink.redis.address=localhost:6379
+
+# Connector
```
smiklosovic, line 54: I do not think this whole Cassandra section is relevant anymore after we switched to Cassandra Java connector of Datastax of version 4. Whole Cassandra related configuration is in a special file driver reacts on and here it is just a property which references that file.

ggaborg:
@smiklosovic Hi, I'm not sure I follow. What I've added is some info related to Debezium Server only. Should I move it to the debezium documentation? Or what you're saying is that the java properties are not needed anymore for running? For Cassandra 3 they are still neded, no? Thanks.

smiklosovic:
> @smiklosovic Hi, I'm not sure I follow. What I've added is some info related to Debezium Server only. Should I move it to the debezium documentation? Or what you're saying is that the java properties are not needed anymore for running? For Cassandra 3 they are still neded, no? Thanks.

@ggaborg Cassandra driver of version 4 is used for both Cassandra 3 as well as Cassandra 4 connectors. If you look into dependencies of the main pom.xml (not core, top-level one), there we use driver of version 4. This was rewritten some time ago as we were splitting the connector between c3 / c4 and core.

So yes, related Cassandra properties are not needed anymore, maybe this discussion is handy for you to follow to see it in action:

https://github.com/debezium/debezium-connector-cassandra/pull/43#issuecomment-1039294501

There is "cassandra.driver.config.file=/opt/debezium/application.conf" entry in the config file and there is also example of that config file. So yeah, most probably the debezium documentation on the web page is obsolete.

core/src/main/java/io/debezium/connector/cassandra/CassandraFactoryImpl.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.cassandra;
+
+import org.apache.kafka.clients.producer.KafkaProducer;
+
+public class CassandraFactoryImpl implements CassandraFactory {
```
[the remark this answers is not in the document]
ggaborg: @smiklosovic MainComponentFactory is ok?

core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorFactory.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.cassandra;
+
+import io.debezium.config.CommonConnectorConfig;
+import io.debezium.connector.base.ChangeEventQueue;
+import io.debezium.pipeline.DataChangeEvent;
+
+public class CassandraConnectorFactory implements CassandraFactory {
```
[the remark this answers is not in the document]
ggaborg: @smiklosovic DebeziumComponentFactory is ok?

core/src/main/java/io/debezium/connector/cassandra/AbstractConnectorTask.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public String version() {
+        return Module.version();
+    }
+
+    @Override
+    protected ChangeEventSourceCoordinator<CassandraPartition, CassandraOffsetContext> start(Configuration config) {
+        CassandraConnectorConfig connectorConfig = new CassandraConnectorConfig(config);
+
+        if (connectorConfig.numOfChangeEventQueues() != 1) {
```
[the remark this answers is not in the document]
ggaborg: @smiklosovic, this is not a validation per-se. the configuration is the same for the connector no matter if we start it standalone or as a debezium connector under debezium server. But for the second case the value of this specific property must be 1. I tried to do the smallest changes possible in order not to affect anything in the standalone connector.

ggaborg:
@smiklosovic I tried to run without all the java properties and it is not working with java 11:
2022-11-22 17:18:25,902 ERROR [org.apa.cas.io.uti.FileUtils] (pool-7-thread-1) FATAL: Cassandra is unable to access required classes. This usually means it has been run without the aid of the standard startup scripts or the scripts have been edited. If this was intentional, and you are attempting to use Java 11+ you may need to add the --add-exports and --add-opens jvm options from either jvm11-server.options or jvm11-client.options: java.lang.IllegalAccessException: access to public member failed: sun.nio.ch.DirectBuffer.cleaner[Ljava.lang.Object;@d24752f/invokeInterface, from class org.apache.cassandra.io.util.FileUtils (unnamed module @26933859)

ahmedjami:
Hey @ggaborg 

The way to launch DBZ with java 11 is described by @smiklosovic  here: https://github.com/debezium/debezium-connector-cassandra/pull/43#issuecomment-1039294501

Hope this helps
Cheers,
Ahmed

core/src/main/java/io/debezium/connector/cassandra/AbstractConnectorTask.java
```diff
@@ -0,0 +1,102 @@
...
+            throw new CassandraConnectorConfigException(String.format("configuration property %s must be equal to 1",
```
smiklosovic, line 47: @ggaborg you might explain this in this log message, explicitly mentioning that it has to be 1 under debezium server

smiklosovic:
@ggaborg btw while you are at it, it would be nice to refactor how we are doing tests when it comes to Kafka. Because if you take a closer look, I do not think we actually need a Kafka started, tests should work without that as well. We are just "mocking" Kafka if I am right ...

https://github.com/debezium/debezium-connector-cassandra/blob/main/core/src/test/java/io/debezium/connector/cassandra/CassandraConnectorTestBase.java#L95-L101

I dont think that is really necessary, it should also run tests way faster. We can limit Kafka start to only specific tests if they really require it and we can not avoid that.

core/src/main/java/io/debezium/connector/cassandra/AbstractConnectorTask.java
ggaborg: @smiklosovic this entire class only applies to debezium, not to standalone

ggaborg:
@smiklosovic That is what I did in regards to the tests, kafka is not needed at all, so I removed it.

smiklosovic:
@ggaborg ah great, I have not noticed that. Thanks!

core/src/main/java/io/debezium/connector/cassandra/AbstractConnectorTask.java
smiklosovic: maybe I interpreted your "_ But for the second case the value of this specific property must be 1_ " incorrectly. What did you mean by "second case"?

smiklosovic:
@ggaborg @jpechane in general it look good to me, I put comments where I could. It would be nice to address naming of some classes to better readability.

One of our testers is going to build this and evaluate if it is indeed a drop-in replacement for 2.0.0 so we have nothing to do when it comes to upgrade in the future.

It would be very nice if this branch is rebased against the current trunk as it is one commit behind but no stress as we can rebase that on our own anyway.

core/src/main/java/io/debezium/connector/cassandra/AbstractConnectorTask.java
ggaborg: @smiklosovic By 'second case' I meant running the connector under Debezium Server, not as standalone. This entire AbstractConnectorTask only applies when running under Debezium Server, not as standalone.

ggaborg:
@smiklosovic @jpechane Rebased, thanks.

jpechane:
@smiklosovic Thanks for the offer for testing! We aim for 2.1.0.Fnal release in the mid-December. Do you have any timeframe when the test results will be available?

smiklosovic:
@jpechane when do you think 2.1.0.Final will be released? We would like to see 2.0.1 with DBZ-5826 in. I am not sure what happens first and if you plan to do some patch releases for 2.0.0 as well.

jpechane:
@smiklosovic Hi, we will do 2.0.1.Final release the next week. 2.1.0.Final should go out around Dec 20th.

jfleming-ic:
Hi, I work with @smiklosovic 

We've tested our use case against this, everything seemed fine.

Looks good on our side 👍

jpechane:
Applied, thanks a lot guys for the smooth cooperation!

PR #77, opened 2022-10-12
DBZ-5719 cassandra connector first startup ever may fail

ggaborg:


jpechane:
@ggaborg Applied, definitely safer solution, thanks!

PR #81, opened 2022-11-11
DBZ-5826 fix names of range fields in schema to comply with Avro standard

smiklosovic:


smiklosovic:
@jpechane kindly pinging

jpechane:
@smiklosovic There is a class `io.debezium.data.VerifyRecord`. It might be useful to place few calls around the codebase to make sure that messages are valid.

smiklosovic:
Is this still necessary to do? I pinged you in the ticket that this might go in as is.

jpechane:
@smiklosovic Applied, thanks! The remark was a generic one as using those methods would catch such issues right at the start.

PR #84, opened 2022-12-08
DBZ-5912 Replace simple string for range tombstones for JSON object

smiklosovic:


smiklosovic:
@jpechane would you take a look?

cc: @ahmedjami

ahmedjami:
Hi @smiklosovic 
Thanks a lot for this PR :) My colleague @JeremyVigny will try it and keep you informed of progress.
Cheers,
Ahmed

core/src/main/java/io/debezium/connector/cassandra/RangeData.java
```diff
@@ -0,0 +1,88 @@
...
+                    .put("method", method)
+                    .put("values", values);
+        }
+        catch (DataException e) {
+            throw new DebeziumException(e);
+        }
+    }
+
+    static Schema rangeSchema(String name) {
+        return SchemaBuilder.struct().name(name)
```
jpechane, line 54: Could you pleasee add all new schemas into `CassandraSchemaFactory` and add an explicit version to them?

JeremyVigny:
Hi @smiklosovic ,

Thanks for this fix. 
Just one last thing. Is it possible to keep the type of values? 
For example: 
`"values: { "sp": 0 }"` instead of: `"values: { "sp": "0" }"`  Because it's an integer.
Thanks, 
Jérémy

core/src/main/java/io/debezium/connector/cassandra/RangeData.java
```diff
@@ -0,0 +1,88 @@
...
+                .field("method", Schema.STRING_SCHEMA)
+                .field("values", SchemaBuilder.map(Schema.STRING_SCHEMA, Schema.STRING_SCHEMA))
```
jpechane, line 56: IMHO the schema here is not correct, you still need to call build after the `map`. Again symobilc name and version number would preferred and created in the schema factory centralized class.

jpechane (changes requested):

jpechane:
@smiklosovic Thanks a lot for PR. It generally LGTM, I just asked for change in schema building.

core/src/main/java/io/debezium/connector/cassandra/RangeData.java
```diff
@@ -0,0 +1,88 @@
...
+    public RangeData(String name, String method, Map<String, String> values) {
+        this.name = name;
+        this.method = method;
+        if (values != null) {
+            this.values.putAll(values);
+        }
+    }
+
+    public static RangeData start(String method, Map<String, String> values) {
+        return new RangeData("_range_start", method, values);
```
jpechane, line 34: Please use constants where possible - here and in the following fields too.

core/src/main/java/io/debezium/connector/cassandra/RangeData.java
smiklosovic: @jpechane could you point me to example of these versions and names? I am not completely sure I follow here.

core/src/main/java/io/debezium/connector/cassandra/RangeData.java
smiklosovic: @jpechane could you point me to example of these versions and names? I am not completely sure I follow here.

smiklosovic: ah ok I see what you mean

cassandra-3/src/test/java/io/debezium/connector/cassandra/RangeTombstoneCommitLogProcessingTest.java
```diff
@@ -67,7 +70,25 @@ public void verifyEvents() throws Exception {
         Record range1 = (Record) events.get(2);
         assertEquals(range1.getEventType(), CHANGE_EVENT);
         assertEquals(RANGE_TOMBSTONE, range1.getOp());
-        assertEquals("INCL_START_BOUND(1)", range1.getRowData().getStart());
-        assertEquals("INCL_END_BOUND(1, 2)", range1.getRowData().getEnd());
+
+        RangeData startRange = range1.getRowData().getStartRange();
+        RangeData endRange = range1.getRowData().getEndRange();
+
+        Assert.assertNotNull(startRange);
```
jpechane, line 77: `assertNotNull` already imported statically.

cassandra-4/src/test/java/io/debezium/connector/cassandra/RangeTombstoneCommitLogProcessingTest.java
```diff
@@ -65,7 +70,25 @@ public void verifyEvents() throws Exception {
         Record range1 = (Record) events.get(2);
         assertEquals(range1.getEventType(), CHANGE_EVENT);
         assertEquals(RANGE_TOMBSTONE, range1.getOp());
-        assertEquals("INCL_START_BOUND(1)", range1.getRowData().getStart());
-        assertEquals("INCL_END_BOUND(1, 2)", range1.getRowData().getEnd());
+
+        RangeData startRange = range1.getRowData().getStartRange();
+        RangeData endRange = range1.getRowData().getEndRange();
+
+        Assert.assertNotNull(startRange);
```
jpechane, line 77: `assertNotNull` already imported statically.

smiklosovic:
@jpechane I have moved everything to `CassandraSchemaFactory`. It does not make sense to move just RangeData there and leave RowData and CellData outside so that makes the patch a little big bigger.

smiklosovic:
@JeremyVigny I am afraid that is not possible. The map has to have some schema where keys and values have to be of this schema, basically, this is taken from `org.apache.kafka.connect.dataSchema`

````
    Schema INT8_SCHEMA = SchemaBuilder.int8().build();
    Schema INT16_SCHEMA = SchemaBuilder.int16().build();
    Schema INT32_SCHEMA = SchemaBuilder.int32().build();
    Schema INT64_SCHEMA = SchemaBuilder.int64().build();
    Schema FLOAT32_SCHEMA = SchemaBuilder.float32().build();
    Schema FLOAT64_SCHEMA = SchemaBuilder.float64().build();
    Schema BOOLEAN_SCHEMA = SchemaBuilder.bool().build();
    Schema STRING_SCHEMA = SchemaBuilder.string().build();
    Schema BYTES_SCHEMA = SchemaBuilder.bytes().build();

    Schema OPTIONAL_INT8_SCHEMA = SchemaBuilder.int8().optional().build();
    Schema OPTIONAL_INT16_SCHEMA = SchemaBuilder.int16().optional().build();
    Schema OPTIONAL_INT32_SCHEMA = SchemaBuilder.int32().optional().build();
    Schema OPTIONAL_INT64_SCHEMA = SchemaBuilder.int64().optional().build();
    Schema OPTIONAL_FLOAT32_SCHEMA = SchemaBuilder.float32().optional().build();
    Schema OPTIONAL_FLOAT64_SCHEMA = SchemaBuilder.float64().optional().build();
    Schema OPTIONAL_BOOLEAN_SCHEMA = SchemaBuilder.bool().optional().build();
    Schema OPTIONAL_STRING_SCHEMA = SchemaBuilder.string().optional().build();
    Schema OPTIONAL_BYTES_SCHEMA = SchemaBuilder.bytes().optional().build();
````

**Comment by smiklosovic** [comment]
@jpechane I see that changes in 5044 basically just moved `Schema` to _factories_ but I just moved everything there. I still consider this approach to be better than just moving `Schema`s there. Why not? It is better to have all implementations of `KafkaRecord` at once place instead of all over the project. I have also made constructors of these implementations private so they can not be created by any other way but going through `CassandraSchemaFactory`.

**Comment by smiklosovic** [comment]
@JeremyVigny 

I have an idea how to pass type there as well, it might be done like this:

What do you think?

````
    "_range_start": {
        "method": "EXCL_START_BOUND",
        "values": [
            {
                "name": "sp",
                "value": "1",
                "type": "org.apache.cassandra.db.marshal.Int32Type"
            },
            {
                "name": "d",
                "value": "12, 23, toto",
                "type": " org.apache.cassandra.db.marshal.UTF8Type"
            },
            {
                "name": "dp",
                "value": "1",
                "type": "org.apache.cassandra.db.marshal.Int32Type"
            },
        ]
    },
}
````

**Comment by jpechane** [comment]
@smiklosovic I agree with your decision on schema factory changes. It inflates the PR but it is in isolated commit so no issue for reviewing.

Adding type as the string is good solution. It is probably the only way how to get to "dynamic" typing.

Could you please check the test results? When fixed I hink I can merge the PR.

**Comment by smiklosovic** [comment]
@jpechane I was struggling a little bit to get the schema right. Would you mind to double check?

@JeremyVigny would you please check this too?

Afterwards I would just merge all three commits as they are. Making the last one part of the first one is little bit tricky because of the middle one as stuff has changed since then. I can indeed spent more time on making it two commits only but that seems to me like unnecessary exercise.

**Comment by JeremyVigny** [comment]
Hi @smiklosovic 
Adding the type is a good idea
I checked and it's ok for me!
Thanks

File: core/src/main/java/io/debezium/connector/cassandra/CassandraSchemaFactory.java [hunk 6]
```diff
@@ -18,4 +49,380 @@ public CassandraSchemaFactory() {
...
+            for (ColumnMetadata cm : tm.getColumns().values()) {
+                columnNames.add(cm.getName().toString());
+                columnTypes.add(cm.getType());
+            }
+
+            return rowSchema(columnNames, columnTypes);
+        }
+
+        static Schema rowSchema(List<String> columnNames, List<DataType> columnsTypes) {
+            SchemaBuilder schemaBuilder = SchemaBuilder.struct().name(Record.AFTER).version(1);
```
jpechane, line 161: @smiklosovic I am sorry for the late reaction. There is a small thing to refactor if possible. Could you please extract the version numbers into constants in the same way as is done in `SchemaFactory` class? So in general the constant definitions will always be schema name followed by schema version separated by en empty line and ideally with a comment added if there are logical groups of schemas.
This would help in future with schema management and maintenance. The code functionality is otherwise good.
Thanks for understaning.

smiklosovic: @jpechane yes I will take a look.

smiklosovic:
@JeremyVigny  I have extracted some constants here

https://github.com/debezium/debezium-connector-cassandra/commit/ec9e889ac36912bd84d4258ce9a5c8d79a3fb844

You may try one more time to be sure all is fine.

PR #85, opened 2022-12-12
DBZ-5923 Support Cassandra TimeUUID values to be encoded as strings

smiklosovic:


jpechane:
@smiklosovic Applied, thanks!

PR #86, opened 2022-12-12
DBZ-5776 enqueue EOF event for log file only on successful processing or on error for Cassandra 4 implementation

smiklosovic:
@ggaborg maybe something to verify that it works for you well?

cc @jpechane


ggaborg:
@smiklosovic Works fine, thanks!

jpechane:
@smiklosovic Applied, thanks!

PR #88, opened 2022-12-13
DBZ-5933 set properties to Cassandra container to make tests faster

smiklosovic:
There are properties in Cassandra we can tweak so startup of a node is faster hence tests are faster. As single-node cluster does not need to wait for other nodes to get schema etc we can just set these properties pretty low.

### DIFF
--- a/core/src/test/java/io/debezium/connector/cassandra/CassandraConnectorTestBase.java
+++ b/core/src/test/java/io/debezium/connector/cassandra/CassandraConnectorTestBase.java
@@ -52,7 +52,8 @@ public abstract class CassandraConnectorTestBase {
             .withExposedPorts(9042)
             .withStartupTimeout(Duration.ofMinutes(3))
             .withCreateContainerCmdModifier(cmd)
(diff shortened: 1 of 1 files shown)


core/src/test/java/io/debezium/connector/cassandra/CassandraConnectorTestBase.java
```diff
@@ -52,7 +52,8 @@ public abstract class CassandraConnectorTestBase {
             .withExposedPorts(9042)
             .withStartupTimeout(Duration.ofMinutes(3))
             .withCreateContainerCmdModifier(cmd)
-            .withFileSystemBind(cassandraDir, CASSANDRA_SERVER_DIR, BindMode.READ_WRITE);
+            .withFileSystemBind(cassandraDir, CASSANDRA_SERVER_DIR, BindMode.READ_WRITE)
+            .withCommand("-Dcassandra.ring_delay_ms=5000 -Dcassandra.superuser_setup_delay_ms=1000");
```
smiklosovic, line 56: default is 30s and 10s respectively

vjuranek (approved):

PR #90, opened 2022-12-14
DBZ-5925 package 'standalone' connector tar/zip archive in assembly profile

keriharris:


keriharris:
This PR is a POC used to help ongoing discussion in https://issues.redhat.com/browse/DBZ-5925. Until there is consensus in DBZ-5925 it should be considered only tentative.

cassandra-4/pom.xml
```diff
@@ -265,6 +265,21 @@
                                     <tarLongFileMode>posix</tarLongFileMode>
                                 </configuration>
                             </execution>
+                            <execution>
```
jpechane, line 268: @keriharris @smiklosovic 
How about using `maven-shade-plugin`? The fatjar can be created with it but it provides an option to exclude a set of dependencies. That way we can get both fatjar with the necessary dependencies and the `tar.gz` as well. WDYT?

smiklosovic: @jpechane worth to investigate and compare the results!

PR #93, opened 2022-12-21
DBZ-5925 Keep fatjar and provide slimmed plugin distro

jpechane:


ggaborg (commented):
same changes in cassandra-3 pom need to be done to cassandra-4 pom

jpechane:
@ggaborg of course, this is just a validation of approach, not even all dpes has been exluded yet. I am trying to find a common ground and chekcing if it is the case.

smiklosovic:
If I understand it right, if debezium-server will contain Cassandra, will server depend - Maven-wise - on what we ship? (server will reference this connector in its dependencies? What does "out of the box mean" here?

Could not we wait until debezium-server is out so we do not need to ship plugin so this whole work would be probably just abandoned?

jpechane:
Closing in favour of https://github.com/debezium/debezium-connector-cassandra/pull/99 and https://github.com/debezium/debezium/pull/4231

PR #95, opened 2022-12-26
DBZ-5950 Move common code in Cassandra connector core module

ggaborg:


ggaborg:
@jpechane @smiklosovic Hi and happy new year! When you have the time, please review this. It is needed for the datastax connector. Thanks

smiklosovic:
@ggaborg this is a great initiative but ... 

If you look at this (1), `core` currently depends on cassandra-3.11. The thing is, and I had this in my mind for a long time, `core` should not depend on _anything_ Cassandra-related. core should be totally generic. Think about that, currently, we are depending, in runtime, on what is in 3.11. I do no think that is safe. What if there is some breaking change in 4.x which will not work in runtime with 3.11 API anymore? I think it is only matter of time until we hit that issue.

So what you did is definitely a good thing. But we might probably go one step further here and investigate what it would take to get rid of that dependency completely.  

(1) https://github.com/debezium/debezium-connector-cassandra/blob/main/core/pom.xml#L17-L23

ggaborg:
@smiklosovic @jpechane cassandra lib dependency from core module is removed now

smiklosovic:
great stuff @ggaborg , I will review this in very near future.

smiklosovic:
@ggaborg I opened this PR (1) against your branch with some cleanup. I think if you merge that it will be automatically reflected here. Also, could you please rebase on top of the current upstream's main?

(1) https://github.com/ggaborg/debezium-connector-cassandra/pull/1

smiklosovic:
@ggaborg could you please also squash it into one commit?

jpechane:
@ggaborg Please no merge commits in the PR

smiklosovic:
> @ggaborg Please no merge commits in the PR

yep, that is why I ask for a squash.

ggaborg:
@jpechane @smiklosovic Hi, can this be merged? Thanks!

smiklosovic:
It looks good to me, I ll leave the pleasure of pushing the button to @jpechane, maybe he wants to take a quick look too?

jpechane:
Rubberstamped and applied :-). @ggaborg Thanks a lot!

PR #98, opened 2023-01-18
DBZ-5951 Create a Datastax connector based on Cassandra connector

ggaborg:


ggaborg:
@smiklosovic, @jpechane Hi, I did not include it as a sub-module in the main Cassandra pom.xml because it has some dependencies which are not available in the maven repo. They need to be installed manually in the local maven repo. It's all described in the README.md. It should be eventually added to the github build flow. Take your time and big thanks!

ggaborg:
Of course, no rush.

smiklosovic:
I have not tried that as I do not have DSE distribution locally and I dont feel like registering / downloading it. I do not have any reason not to believe it works as it is in Gabor's interests. I would like to comment more generally about what I see.

1) the fact that this is not part of the build is quite unfortunate. We can not ship it and we can not build it in CI either. I do not know what to do with this situation. We can not even put it among modules and somehow change the CI in such a way that it would be skipped because the mere fact that it would be a module will mark it with errors in IDE(A) as the dependencies are not there. So, out of the box, when somebody imports the project, it will be failing without them going to Datastax and downloading DSE. So we would make this dependable on Datastax which is not good, obviously.

2) When this is not part of the build, we do not know if it works. Committers would have to be the ones who have local DSE available just for the sake of trying it out.

3) We are starting to be too repetitive in tests. If there is Cassandra 5 release later this year (hopefully), we would have fourth set of tests? This is becoming hard to manage.

I was thinking what approach we might take and what we could do is to start these tests programmatically. We would have one class in tests for each module which would programmatically call all tests which would be located in the project just once. The problem is that individual tests are extending `AbstractCommitLogProcessorTest` which in turn extends `DseConnectorTestBase`. It is similarly done for all other modules. Maybe we could rewrite tests in such a way that this would be dynamically loaded, e.g. via Java SPI mechanism when that test is being executed. (2)

(1) https://www.baeldung.com/junit-tests-run-programmatically-from-java
(2) https://www.baeldung.com/java-spi

I do no say this should not end up being merged, not at all. I think it is very good and important contribution. There are some caveats along the way we need to solve / agree to.

jpechane:
@ggaborg @smiklosovic Thanks guys for the contribution and input.

- [ ]  Please make the new module parent of the main cassandra pom but make it build only if an explicit profile is enabled
- [ ]  Provide a bash script that will automatically set-up/provide dependencies to make the connector budildable. The missing deps are located in the Docker image used for testing so the script will pull the images, extract JARs amd isntall them to the local Maven repo
- [ ] Modify CI job to run the script and execut build with the new DataStax profile

Regarding tests, I beleive it is really good point. I think it should not be handled as a part of this PR but a new Jira should be created for that and a protype tried later after this PR is merged.

Thoughts?

smiklosovic:
I am fine with taking care of the tests later.

ggaborg:
@jpechane Can I use docker command in the bash script?

jpechane:
@ggaborg Yes, I think it should be feasible

ggaborg:
@jpechane done

smiklosovic:
@ggaborg `install-artifacts.sh` should be `chmod +x`-ed to be executable and I would move it to the root of the repository to be clearly visible to be executed to setup the project.

dse/install-artifacts.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+DSE_VERSION=6.8.16
+mvn dependency:get -Dartifact=com.datastax:dse-db-all:${DSE_VERSION} -o && mvn dependency:get -Dartifact=com.datastax:dse-commons:${DSE_VERSION} -o && mvn dependency:get -Dartifact=io.netty:netty-all:4.1.25.7.dse
```
smiklosovic, line 4: could we maybe redirect error output to dev null or something like that? What I see locally is that it will try to get this and if it is not OK, it will download image etc. What I kind of do not like to see is that error but that is just a personal preference ....

dse/pom.xml
```diff
@@ -0,0 +1,340 @@
...
+    <!--
+    Define several useful profiles
+    -->
+    <profiles>
+        <profile>
+            <id>assembly</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <dependencies>
```
smiklosovic, line 163: I don't think all these dependencies are necessary. Please check how it is done in current main branch. This was reworked like a week ago so you probably haven't caught the changes yet ...

smiklosovic:
@ggaborg are you still around? If not, we should probably take it from here? cc @jpechane

smiklosovic:
Thank you for letting us know, @ggaborg , I opened this PR with few minor changes otherwise all seems to be fine. I ll finish it there. Thanks!

(1) https://github.com/debezium/debezium-connector-cassandra/pull/106

jpechane:
@smiklosovic @ggaborg I am sorry for the super late reply, I was a bit buried last few days. Please expect my feedback ro merge no later than next Tue. Thanks for understanding,

jpechane:
Merged via https://github.com/debezium/debezium-connector-cassandra/pull/106

PR #104, opened 2023-02-06
DBZ-6076: Add support for custom SourceInfoStructMaker for adding new fields to source field

adasari:


smiklosovic:
@adasari 

Where are you getting that from?

https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java

There is 

`BaseSourceTask<P extends Partition, O extends OffsetContext>`

Not the version you suggested.

If you want this to happen, you need to make that change in Debezim core first and the return to this ticket.

PR #110, opened 2023-04-10
DBZ-6320 Fix Cassandra3 build to work with JDK names sans dots.

Naros:
https://issues.redhat.com/browse/DBZ-6320

The most recent JDK 20 version string is simply "20" and the jamm dependency bundled with cassandra-all does not support version strings that may not contain dots, basically it only supports JDK 8 or earlier. Bumping to 0.3.3 supports dot and non-dot named version strings.

@jpechane This should fix the Java Outreach failures with JDK20 for Cassandra builds.

### DIFF
--- a/cassandra-3/pom.xml
+++ b/cassandra-3/pom.xml
@@ -13,6 +13,21 @@
     <name>Debezium Connector for Cassandra 3</name>
     <packaging>jar</packaging>
 
+    <properties>
+        <!-- Fixes compatibility JDK versions that may not yet dots in their names -->
+        <version.jamm>0.3.3</version.jamm>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+        <dependency>
+            <groupId>com.github.jbellis</groupId>
+            <artifactId>jamm</artifactId>
+            <version>${version.jamm}</version>
+        </dependency>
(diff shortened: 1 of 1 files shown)


smiklosovic:
what about cassandra 4 and dse module? All good there?

cassandra-3/pom.xml
```diff
@@ -45,8 +45,19 @@
...
+                    <groupId>com.github.jbellis</groupId>
+                    <artifactId>jamm</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
+        <!-- Fixes compatibility JDK versions that may not yet dots in their names -->
+        <dependency>
+            <groupId>com.github.jbellis</groupId>
+            <artifactId>jamm</artifactId>
+            <version>0.3.3</version>
```
jpechane, line 59: Should not this go into BOM instead?

Naros: I choose not to do that because the only one I wanted to override was cassandra 3 as the other two modules are unaffected.  If you want it defined in the BOM I can do that, just let me know.

jpechane:
@Naros Applied, thanks. I just changed the POM to use dependency management for version resolution.

PR #118, opened 2024-01-18
DBZ-7363 Debezium stops producing events on schema changes in cassandra4

schampilomatis:
Adjust cassandra4 schema listener to use old table metadata Table ID when replacing the metadata in Cassandra 4 Schema Provider. Replace current logic of calculating the uuid from keyspace name and table name which does not match.


schampilomatis:
@jpechane I updated the branch with formatting, can you retrigger the workflow?

jpechane:
@schampilomatis Applied, thanks a lot!

PR #124, opened 2024-03-11
DBZ-7618 Prevent multiple calls to pu.maxTimestamp()

samssh:
Invoking the `pu.maxTimestamp()` function within the `handleRowModifications` method, which is inside a loop in `handleRowIterator`, can lead to performance issues and significantly slow down the connector.

### DIFF
--- a/cassandra-3/src/main/java/io/debezium/connector/cassandra/Cassandra3CommitLogReadHandlerImpl.java
+++ b/cassandra-3/src/main/java/io/debezium/connector/cassandra/Cassandra3CommitLogReadHandlerImpl.java
@@ -308,6 +308,7 @@ public class Cassandra3CommitLogReadHandlerImpl implements CommitLogReadHandler 
     }
 
     private void handleRowIterator(PartitionUpdate pu, OffsetPosition offsetPosition, KeyspaceTable keyspaceTable) {
+        long maxTimePu = pu.maxTimestamp();
         UnfilteredRowIterator it = pu.unfilteredIterator();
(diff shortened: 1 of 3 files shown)


samssh:
https://issues.redhat.com/browse/DBZ-7622

jpechane:
@samssh Applied, thanks!

Ficooooooo:
> Invoking the `pu.maxTimestamp()` function within the `handleRowModifications` method, which is inside a loop in `handleRowIterator`, can lead to performance issues and significantly slow down the connector.

Hi @samssh,  I've encountered with performance issue as well, and I've test performance of 2.5.3.Final, it seems like there is no difference on performance when reading the commit log files, did I use the wrong version or I need to change some configurations? 

And based on your test, is the problem fixed?

samssh:
> > Invoking the `pu.maxTimestamp()` function within the `handleRowModifications` method, which is inside a loop in `handleRowIterator`, can lead to performance issues and significantly slow down the connector.
> 
> Hi @samssh, I've encountered with performance issue as well, and I've test performance of 2.5.3.Final, it seems like there is no difference on performance when reading the commit log files, did I use the wrong version or I need to change some configurations?
> 
> And based on your test, is the problem fixed?

Hi @Ficooooooo. I answered your question [here](https://issues.redhat.com/browse/DBZ-7704?focusedId=24466621&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24466621)

PR #126, opened 2024-03-12
DBZ-7631 Change hashing strategy to use partition columns

samssh:
To ensure the proper ordering of events within the same partition, it's necessary to utilize the partition columns when determining the queue index for event insertion.

### DIFF
--- a/cassandra-3/src/main/java/io/debezium/connector/cassandra/Cassandra3CommitLogReadHandlerImpl.java
+++ b/cassandra-3/src/main/java/io/debezium/connector/cassandra/Cassandra3CommitLogReadHandlerImpl.java
@@ -81,6 +81,7 @@ public class Cassandra3CommitLogReadHandlerImpl implements CommitLogReadHandler 
     private final CommitLogProcessorMetrics metrics;
     private final RangeTombstoneContext<CFMetaData> rangeTombstoneContext = new RangeTombstoneContext<>();
     private final CassandraSchemaFactory schemaFactory;
+    private final CassandraConnectorConfig.EventOrderGuaranteeMode eventOrderGuaranteeMode;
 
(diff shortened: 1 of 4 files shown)


samssh:
https://issues.redhat.com/browse/DBZ-7631

cassandra-3/src/main/java/io/debezium/connector/cassandra/Cassandra3CommitLogReadHandlerImpl.java
```diff
@@ -665,4 +665,8 @@ private List<Object> getPartitionKeys(PartitionUpdate pu) {
         return values;
     }
 
+    private int getPartitionQueueIndex(PartitionUpdate partitionUpdate) {
```
jpechane, line 668: Can this method be extracted into a utility class in core module and re-used in all three modules?

jpechane:
@samssh Thank you for the PR. I left a comment related to a small refactoring to re-use the code.
Still this is quite breaking change and I can imagine other users prefering current behaviour. Would it be possible to create a configuration option that would control the hashing startegy? Ideally enum with two options that would allow additonal expansion if needed.

samssh:
You're welcome.

Regarding the method refactoring, unfortunately, it's not possible to extract the method to a utility class in the core module because for every version of Cassandra we have a different artifact that contains the `PartitionUpdate` class. If you agree, I can create an abstraction and develop a wrapper class to move more common logic to the core module, but this will likely require significant changes to the codebase.




And regarding configuring the behavior, that's fine, and I will add a configuration property with an enum as soon as possible.

jpechane:
@samssh I see, that's unfortunate. Let's forget about the refactoring then.

samssh:
@jpechane It should be ready.

samssh:
Thank you. Of course! I'll get on it as soon as possible.

PR #136, opened 2024-10-29
DBZ-8363 creation of single test artifact

smiklosovic:
@jpechane 

This PR introduces a single test artifact which is made a dependency of each respective implementation module. By doing that, we run all tests (when not excluded programmatically) for each module while tests are written just once.

The solution uses ServiceLoader mechanism for loading bits which differ and tests load this and use it.

For review purposes, it is adviced if reviewers open the code in IDE to see what is going on here.

From development perspective, the workflow is that when a new test is being written, it would be written in respective module but upon commit, it would need to be moved into tests artifact. Upon development, a developer will very likely want to have a quick feedback loop and placing test into tests artifact means he is not able to test that individual test in IDE. Hence, a test would be developed in some module but it would be moved to tests upon commit. Another option is to run just one test via Maven "somehow" but a developer very likely wants to debug in IDE as well, hence the first approach is suggested.

If that test class is meant to be run just for a specific Cassandra version, it might stay in the implementation module instead of the tests module.


core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -347,11 +347,22 @@ public static EventOrderGuaranteeMode parse(String value) {
...
+            .withType(Type.STRING)
+            .withDescription("Name of Cassandra cluster.")
+            .withDefault("Test Cluster");
+
     public static final Field COMMIT_LOG_RELOCATION_DIR = Field.create("commit.log.relocation.dir")
             .withType(Type.STRING)
             .withValidation(Field::isRequired)
             .withDescription("The local directory which commit logs get relocated to once processed.");
 
+    public static final Field CASSANDRA_CDC_LOG_LOCATION = Field.create("commit.log.location")
```
jpechane, line 360: Is this used somewhere else than in tests? If not then it should not be exposed as connector config option.

core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -347,11 +347,22 @@ public static EventOrderGuaranteeMode parse(String value) {
             .withType(Type.STRING)
             .withDescription("The absolute path of the YAML config file used by a Cassandra node.");
 
+    public static final Field CASSANDRA_CLUSTER_NAME = Field.create("cassandra.cluster.name")
+            .withType(Type.STRING)
+            .withDescription("Name of Cassandra cluster.")
+            .withDefault("Test Cluster");
```
jpechane, line 353: Why it was not necessary to override it before?

jpechane (commented):
@smiklosovic Thanks a lot! I really like the approach. Te only comments I do have relates to changes in `CassandraConnectorConfig`. If something new is intorduced there then there should be a separate Jira (enhancement) and separate commit for it. The new config options would need a documentation PR too.

core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -599,6 +610,14 @@ public String cassandraConfig() {
         return this.getConfig().getString(CASSANDRA_CONFIG);
     }
 
+    public String clusterName() {
+        return this.getConfig().getString(CASSANDRA_CLUSTER_NAME);
+    }
+
+    public String getCDCLogLocation() {
```
jpechane, line 617:
```suggestion
    public String getCdcLogLocation() {
```

smiklosovic:
@jpechane thank you for the feedback, I shall get back to this next week.

core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
smiklosovic: It is used in production. In `CassandraConnectorTaskTemplate` in `initProcessorGroup`. Before that, there was `deserializerProvider.getClusterName()` at `taskContext.getClusterName()` place.

````
processorGroup.addProcessor(new SnapshotProcessor(taskContext, taskContext.getClusterName()));
````

Now if you think, about that, what the hell does `CassandraTypeProvider` have in common with name of the cluster? It was implemented like this:

````
    @Override
    public String getClusterName() {
        return DatabaseDescriptor.getClusterName();
    }
````

So, the only reason we put this into `CassandraTypeProvider` before was that we were trying to "hide" this _somewhere_ so `DatabaseDescriptor` is on the class path and it is module specific. We could not put this into "core" because `DatabaseDescriptor` is not there (we are not depending on any specific Cassandra version there). 

Cluster name should be fetched from `CassandraConnectorContext`. We have one implementation of the context - `DefaultCassandraConnectorContext` which is in `core`, in `CassandraConnectorTaskTemplate`. So we can not put there `DatabaseDescriptor.getClusterName` because that class is not in `core`. Hence I made a configuration property out of that.

The same logic holds for the other property.

jpechane: @smiklosovic We've also `createInternal` to create field definition. So it might be fine to use it and just add proper description.

smiklosovic: @jpechane I created https://issues.redhat.com/browse/DBZ-8373

I have removed it and put it back to CassandraTypeProvider even it does not make sense.

smiklosovic: @jpechane ah, I noticed you comment just now ... I am not completely sure if `createInternal` is the right thing to do. It is not _internal_. I think that property (cluster name) which is used in `SnapshotProcessor` uses name of cluster internally etc so I think that property is really something a user should have a control over.

jpechane: So, lte me rephrase it. If it is something that user can and should use then let's keep it as is. I don't like the default value but it is apparently matching the one in driver so I am going to accept it.
BTW, there slight misunderstanding. I did not ask originally for the removal, just explanation.

jpechane:
@smiklosovic I see it built

core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
smiklosovic: @jpechane I am fine to rework it in DBZ-8373 just to be more explicit about that. If it was delivered with this patch then it would be all burried in this big patch. Let's just deliver this as it is now and then we can refactor in the other one. That is totally fine with me.

jpechane:
@smiklosovic Deal done, applied. Thanks!

PR #146, opened 2025-04-14
DBZ-8933 Improve field exclude list validation in FieldFilterSelector

kavyaramaiah:
- Add robust input validation for field exclude list entries
- Implement proper error handling with descriptive messages
- Validate format requirements (keyspace.table.column)
- Add whitespace trimming for all field components
- Prevent empty keyspace/table names
- Fix ArrayIndexOutOfBoundsException when parsing malformed entries

This change improves error reporting and prevents crashes when processing  incorrectly formatted field exclude list configurations in the Debezium  Cassandra connector.

### DIFF
--- a/core/src/main/java/io/debezium/connector/cassandra/FieldFilterSelector.java
+++ b/core/src/main/java/io/debezium/connector/cassandra/FieldFilterSelector.java
@@ -8,6 +8,8 @@ package io.debezium.connector.cassandra;
 import java.util.ArrayList;
 import java.util.List;
 
+import com.google.common.base.Strings;
+
 import io.debezium.connector.cassandra.CassandraSchemaFactory.RowData;
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/debezium/connector/cassandra/FieldFilterSelector.java
```diff
@@ -61,11 +61,28 @@ private static final class Field {
         final String column;
 
         private Field(String fieldExcludeList) {
-            String[] elements = fieldExcludeList.split("\\.", 3);
-            String keyspace = elements[0];
-            String table = elements[1];
+            if (fieldExcludeList == null || fieldExcludeList.trim().isEmpty()) {
```
Naros, line 64:
```suggestion
            if (Strings.isNullOrEmpty(fieldExcludeList)) {
```

core/src/main/java/io/debezium/connector/cassandra/FieldFilterSelector.java
```diff
@@ -61,11 +61,28 @@ private static final class Field {
...
+            String table = elements[1].trim();
+            if (keyspace.isEmpty() || table.isEmpty()) {
+                throw new IllegalArgumentException(
+                    "Keyspace and table names cannot be empty in: '" + fieldExcludeList + "'"
+                );
+            }
+
             keyspaceTable = new KeyspaceTable(keyspace, table);
-            column = elements[2];
+            column = elements[2].trim();
```
Naros, line 85: We check whether `keyspace` and `table` are empty, but we don't do this for `column`. Should we guard against the `column` being empty too?

Naros (changes requested):
A few inline comments.

core/src/main/java/io/debezium/connector/cassandra/FieldFilterSelector.java
kavyaramaiah: Yes, it is good to validate on column as well, I will make the changes

jpechane:
@kavyaramaiah The PR is neither properly formatted nor compilebale. Could you please try to build it locally first and then push the fix? Thanks!

kavyaramaiah:
> @kavyaramaiah The PR is neither properly formatted nor compilebale. Could you please try to build it locally first and then push the fix? Thanks!

Sure, there was some issue while pushing. I will update

kavyaramaiah:
> @kavyaramaiah The PR is neither properly formatted nor compilebale. Could you please try to build it locally first and then push the fix? Thanks!

Please check now

jpechane:
@kavyaramaiah Applied, thanks!

PR #173, opened 2026-03-31
debezium/dbz#1769 Add OpenTelemetry tracing to debezium-connector-cassandra

### LINKED ISSUE #1769: Add OpenTelemetry tracing support to debezium-connector-cassandra
## Feature request or enhancement

_For feature requests or enhancements, please provide the following information:_

**Which use case/requirement will be addressed by the proposed feature?**

The Cassandra connector currently has no tracing support, making it impossible to correlate CDC events with distributed traces across the pipeline. Other Debezium connectors support tracing via the ActivateTracingSpan SMT, but since the Cassandra connector does not use the Kafka Connect framework, that is unavailable.

Users running Cassandra CDC pipelines need end-to-end observability: knowing when a Cassandra mutation occurred, when it was processed by Debezium, and being able to continue the trace downstream (e.g. into an Elasticsearch Sink or other consumers). 

---

**Implementation ideas (optional)**

Instrument `debezium-connector-cassandra` at the Kafka Emitter level, following the same OTEL span conventions used in `debezium-core`. Two nested OTEL spans would be created per emitted event:                                                                                         
  - `db-log-write`: backdated to the Cassandra mutation timestamp (`tsMicro`), with source attributes: `keyspace, table, cluster, commitlog file/position, snapshot flag, connector version`.    
  - `debezium-read`: timestamped at Debezium processing time, with `op` and `ts_ms`.                                                                                                                                                                                                                                                                                            

A trace context is injected into the Kafka ProducerRecord headers (`traceparent`) so downstream consumers can continue the distributed trace.

A new config property `tracing.enabled` (default: false) would gate the feature. If tracing is enabled but no OTEL jar is present on the classpath, the connector no-ops rather than failing (same as `debezium-core`).

jogomez97:
## Summary
Since the Cassandra connector has zero tracing support, this PR instruments the CDC pipeline directly at the Kafka Emitter level, following the same OTEL span conventions as debezium-core.

Two nested OTEL spans are created per each Debezium event emitted:
- `db-log-write`: backdated to the Cassandra mutation timestamp (`tsMicro`), with source attributes: keyspace, table, cluster, commitlog file/position, snapshot flag, connector version
- `debezium-read`: timestamped at Debezium processing time, with `op` and `ts_ms`

A trace context is injected into Kafka `ProducerRecord` headers (`traceparent`) so downstream consumers can continue the distributed trace (e.g. Elasticsearch Sink).

Other Debezium connectors expose tracing via the `ActivateTracingSpan` SMT, but since the Cassandra connector is not based on the Kafka Connect framework that is not supported.

Tracing is enabled via `tracing.enabled=true` in connector config, disabled by default. If tracing is enabled but an OTEL jar is not provided, the connector will no-op instead of failing.

## Testing
New unit tests are included. Tested in Cassandra 5 cluster with Debezium `3.3.0-Final` version.

Fixes https://github.com/debezium/dbz/issues/1769

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -19,6 +19,11 @@
             <artifactId>commons-math3</artifactId>
             <version>3.2</version>
         </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
+++ b/core/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
@@ -357,6 +357,17 @@ public class CassandraConnectorConfig extends CommonConnectorConfig {
             .withValidation(Field::isRequired)
             .withDescription("The local directory which commit logs get relocated to once processed.");
 
+    public static final boolean DEFAULT_TRACING_ENABLED = false;
+    public static final Field TRACING_ENABLED = Field.create("tracing.enabled")
+            .withType(Type.BOOLEAN)
+            .withDefault(DEFAULT_TRACING_ENABLED)
+            .withDescription("Enables OpenTelemetry tracing. When enabled and the OpenTelemetry API "
+                    + "is on the classpath, the connector creates tracing spans for each change event.");
+
+    public boolean isTracingEnabled() {
+        return this.getConfig().getBoolean(TRACING_ENABLED);
+    }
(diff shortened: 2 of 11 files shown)


jpechane:
@jogomez97 Applied, great job! Could you please send a docs PR documenting the new functionality - not only the config option but probably a new section in cassandra docs is required.

jogomez97:
> @jogomez97 Applied, great job! Could you please send a docs PR documenting the new functionality - not only the config option but probably a new section in cassandra docs is required.

Thanks! I've created a documentation PR here: https://github.com/debezium/debezium/pull/7313

PR #174, opened 2026-04-04
debezium/dbz#1791 Fix unbounded task queue in FileOffsetWriter causing heap exhaustion

### LINKED ISSUE #1791: Cassandra connector FileOffsetWriter unbounded task queue causes heap exhaustion
FileOffsetWriter.markOffset() submits a lambda to a single-threaded executor with an unbounded LinkedBlockingQueue for every record produced to Kafka. With the default offset.flush.interval.ms=0 (AlwaysFlushPolicy), each task performs disk I/O. At high throughput the executor cannot keep up, causing millions of FutureTask objects to accumulate in the queue and eventually exhausting the heap.

Symptoms: after ~2 hours of operation, Kafka producer latency degrades from 2ms to 800ms, throughput drops, CDC logs pile up, and restart is required.

nirnx:
Closes debezium/dbz#1791

## Summary

- `FileOffsetWriter.markOffset()` submitted a lambda to a single-threaded executor with an unbounded `LinkedBlockingQueue` for every record produced to Kafka
- With the default `offset.flush.interval.ms=0` (`AlwaysFlushPolicy`), each task performed disk I/O — at ~6,000 records/sec the executor could only drain ~200 tasks/sec
- The queue accumulated millions of `FutureTask` objects over time, causing heap exhaustion, GC pressure, and Kafka producer latency degradation
- Restart temporarily fixed it until the queue filled again (~2 hours)

### Fix

Update the in-memory `Properties` directly on the caller thread (thread-safe since `Properties` extends `Hashtable`) and delegate only the disk flush to the executor. An `AtomicBoolean` guard ensures at most one flush task is ever queued.

### DIFF
--- a/core/src/main/java/io/debezium/connector/cassandra/FileOffsetWriter.java
+++ b/core/src/main/java/io/debezium/connector/cassandra/FileOffsetWriter.java
@@ -17,9 +17,11 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.util.Properties;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


Naros (approved):

jpechane:
@nirnx Applied, thanks!

PR #179, opened 2026-06-01
debezium/dbz#1996 Fix template NPE during startup

Naros:
Fixes debezium/dbz#1996

## Description

Decouples the template instantiation from the factory, allowing for the task context to be accessible earlier to satisfy the requirements inside the new `preStart()` method, avoiding the NullPointerException.


jpechane:
@Naros Any idea why it was not caught with tests?

jpechane (approved):
@Naros LGTM, thanks!

I've added a simple regression test. If it is OK for you feel free to merge yourself.
This also requires 3.5 backport.

Naros:
> @Naros Any idea why it was not caught with tests?

@jpechane could it be because we don't test Debezium Server + Cassandra? At least over in this repository, I believe it's using the basic `static void main` entry point.

PR #182, opened 2026-07-23
debezium/dbz#2273 Bump com.google.guava:guava to 32.0.0

Naros:
Fixes debezium/dbz#2273


vjuranek:
Cassandra3 failure is related to the update.
AI says: `What changed in Guava: The CharMatcher.DIGIT constant field was deprecated in Guava 20.0, and removed in Guava 26.0, replaced by the method CharMatcher.digit().`

Naros:
@vjuranek @jpechane there are a few ways to work around this:
* We override the library back to 24.1.1-jre specifically for Cassandra 3
* We drop support for Cassandra 3, which entered EoL in November 2025.

My personal preference would be to drop Cassandra 3, but curious what you both think.

vjuranek:
Given it seems we cannot replace `CharMatcher.DIGIT` easily as it deep in Cassandra, it really seems the only option is to use different lib for different Cassandra versions, making the maintenance harder. So I'd concur to drop the support for Cassandra 3. If users still need Cassandra 3 support, they can stay on Debezium 3.6.

vjuranek (approved):
LGTM

PR #185, opened 2026-08-10
debezium/dbz#2388 Register all tables in Schema.instance

### LINKED ISSUE #2388: Cassandra connector silently drops cdc=true change events bundled with cdc=false tables in the same Mutation
## Bug report

debezium-connector-cassandra, `3.7.0`

---

Default Cassandra connector configuration with cdc enabled on a subset of tables in a keyspace. Reproduction only requires two tables in the same keyspace, one with `WITH cdc = true` and one with `WITH cdc = false`, sharing a partition key.

---

Reproduced against Cassandra 4.x and Cassandra 5.x (via Testcontainers, local Docker, K8s). 

---

A `cdc=true` table's change event should be emitted regardless of what other tables' PartitionUpdates happen to be bundled into the same Cassandra commit-log Mutation. A `cdc=false` table sharing a batch/partition key with a `cdc=true` table must not cause the `cdc=true` event to be dropped.

---

**What behavior do you see?**

When a single Cassandra Mutation on the commit log bundles PartitionUpdates for two tables that share a partition key, one with `cdc=true `and one with `cdc=false`, the entire Mutation is silently dropped, including the `cdc=true` table's change event. Zero events are emitted for the `cdc=true` table's insert.

Root cause: at startup (`onSessionReady`), the schema-change listener only iterates cdc=true tables (via `getCdcEnabledTableMetadataList`) and registers just those into Cassandra's internal `Schema.instance`. The commit-log deserializer, however, needs every TableId it encounters, regardless of cdc status, to be resolvable via `Schema.instance`, because Mutations are decoded as a whole before any per-table filtering happens. A `cdc=false` table that was never registered causes an `UnknownTableException` on its own PartitionUpdate, and that failure propagates to drop the entire Mutation, including any other tables' PartitionUpdates (`cdc=true` or not) that were bundled into it.

---

Reproduced on `3.7.0-SNAPSHOT` (main branch).

---

Could not find any logs that point to the mutations getting dropped.

---

**How to reproduce the issue?**

Reproduced via the connector's own integration test harness. Steps:

1. Create two tables in the same keyspace sharing a partition key, one cdc-enabled and one not:

```sql
CREATE TABLE test_keyspace.table_a (a int, b int, PRIMARY KEY ((a), b)) WITH cdc = true;
CREATE TABLE test_keyspace.table_b (a int, b int, PRIMARY KEY ((a), b)) WITH cdc = false;
```

2. Write to both tables at the same partition-key value inside a single UNLOGGED BATCH, so the driver/server coalesces both PartitionUpdates into one Mutation on the commit log:

```sql
BEGIN UNLOGGED BATCH
  INSERT INTO test_keyspace.table_a (a, b) VALUES (1, 2);
  INSERT INTO test_keyspace.table_b (a, b) VALUES (1, 20);
APPLY BATCH;
```

3. Observe the connector's change-event output.

   Actual: 0 change events are emitted for table_a's insert (Mutation silently dropped due to UnknownTableException on table_b, which was never registered in Schema.instance because it has cdc=false).
   Expected: 1 change event emitted for table_a's insert (a=1, b=2), regardless of table_b's cdc status.

SreedevT:
## Summary

Fixes debezium/dbz#2388
The Cassandra connector silently drops `cdc=true` change events when their `PartitionUpdate` is bundled in the same `Mutation` as a `cdc=false` table's `PartitionUpdate`.

## Root cause

`onSessionReady` in each schema-change listener only registered `cdc=true` tables into Cassandra's internal `Schema.instance` at startup. When the commit-log deserializer encounters a `Mutation` containing a `PartitionUpdate` for a table that isn't in `Schema.instance`, it throws `UnknownTableException`, and the entire `Mutation` is dropped, including any `cdc=true` `PartitionUpdate` bundled alongside it (e.g. in an `UNLOGGED BATCH` across two tables sharing a partition key).

## Fix

- Added `getAllTableMetadataList` to `AbstractSchemaChangeListener`, which returns all non-virtual tables that have a `cdc` option set.
- `onSessionReady` in all three listeners now bootstraps `Schema.instance` from this full list instead of the cdc-only list, so every table is resolvable regardless of its cdc flag.
- Removed `getCdcEnabledTableMetadataList`, which had no remaining callers after this change.
- Added `MixedCdcBatchSchemaChangeListenerTest`: a regression test with one `cdc=true` and one `cdc=false` table sharing a partition key, batched into a single `UNLOGGED BATCH`, asserting the `cdc=true` insert is still emitted.

## Testing

- `MixedCdcBatchSchemaChangeListenerTest` and `SchemaChangeListenerTest` run against a live Cassandra 5 Testcontainer.
- Verified via logs that both `cdc=true` and `cdc=false` tables are registered into `Schema.instance` at startup, while `schemaHolder`'s "CDC enabled entities" logging remains correctly scoped to `cdc=true` only.


cassandra-4/src/main/java/io/debezium/connector/cassandra/Cassandra4SchemaChangeListener.java
```diff
@@ -43,9 +43,8 @@ public Cassandra4SchemaChangeListener(String kafkaTopicPrefix,
     @Override
     public void onSessionReady(Session session) {
         LOGGER.info("Initializing SchemaHolder ...");
-        List<com.datastax.oss.driver.api.core.metadata.schema.TableMetadata> cdcEnabledTableMetadataList = getCdcEnabledTableMetadataList(session);
-        for (com.datastax.oss.driver.api.core.metadata.schema.TableMetadata tm : cdcEnabledTableMetadataList) {
-            schemaHolder.addOrUpdateTableSchema(new KeyspaceTable(tm), getKeyValueSchema(tm));
```
SreedevT, line 48: onTableCreated() below already calls schemaHolder.addOrUpdateTableSchema() for any table whose cdc option is true, so iterating all tables through it populates SchemaHolder identically while also registering the cdc=false tables into Schema.instance.

vjuranek (approved):
LGTM, but would be good to get at least one more review as I'm not very familiar with Cassandra.

core/src/main/java/io/debezium/connector/cassandra/AbstractSchemaChangeListener.java
```diff
@@ -37,7 +37,10 @@ public AbstractSchemaChangeListener(String kafkaTopicPrefix,
         this.schemaHolder = schemaHolder;
     }
 
-    public List<TableMetadata> getCdcEnabledTableMetadataList(final Session session) {
+    /**
+     * All tables with a cdc option set, virtual and cdc-missing tables excluded.
+     */
+    public List<TableMetadata> getAllTableMetadataList(final Session session) {
```
Naros: I think the `getAll` prefix is a bit misleading here, and I think something like `getTablesWithCdcOption` as the prefix is self-documenting and avoids the need to add the comment you introduced.

core/src/main/java/io/debezium/connector/cassandra/AbstractSchemaChangeListener.java
```diff
@@ -48,13 +51,12 @@ public List<TableMetadata> getCdcEnabledTableMetadataList(final Session session)
                         logger.info(format("Skipping virtual table %s.%s", tm.getKeyspace().asInternal(), tm.getName()));
                         return false;
                     }
-                    Object cdc = tm.getOptions().get(CqlIdentifier.fromCql("cdc"));
-                    if (cdc == null) {
+                    if (tm.getOptions().get(CqlIdentifier.fromCql("cdc")) == null) {
```
Naros, line 51: Semantically, this change is correct, but I am curious what the impact might be for large-scale Cassandra environments with few CDC-enabled tables; how much could this inflate memory consumption as a side effect?

Naros (commented):
Overall LGTM, I've just left one question and one nit-pick but the solution is sound.

core/src/main/java/io/debezium/connector/cassandra/AbstractSchemaChangeListener.java
SreedevT: Yeah that's fair. What about `getCdcOptionedTableMetadataList`?

core/src/main/java/io/debezium/connector/cassandra/AbstractSchemaChangeListener.java
SreedevT: I checked out the memory footprint locally with 200 tables (198 cdc=false, 2cdc=true). The delta between cdc-only and all tables was ~62MB. Most of this came from the TableMetrics/JMX MBeans allocated by the ColumnFamilyStore. We can probably get away with not register CFS for cdc=false tables but I think it would be better as a followup PR.

SreedevT:
@vjuranek @Naros Can this be merged

Naros (approved):

Naros:
Applied, thanks @SreedevT :tada:

PR #187, opened 2026-09-04
debezium/dbz#2541 Fix NPE in onTableUpdated when cdc table option is absent

Thiagoag:
When CDC is enabled on a pre-existing table via `ALTER TABLE ... WITH cdc = true`, the old
table metadata has no `cdc` option, so `getOptions().get(...)` returns null and
`onTableUpdated()` threw a NullPointerException on `toString()`, silently leaving the table
unregistered for CDC. Adds null checks before `toString()` in the Cassandra 4 and Cassandra 5
schema change listeners.


vjuranek (approved):
LGTM, let's wait for CI.

vjuranek:
@Thiagoag  could you please take a look on test failures if they are related to this PR? AFAICT all the tests were passing in the past.

Thiagoag:
> @Thiagoag could you please take a look on test failures if they are related to this PR? AFAICT all the tests were passing in the past.

Hi @vjuranek, the CI failures were from my test, not the fix — it called onTableUpdated(...), which mutates the global Schema.instance. Now it asserts on the extracted isCdcEnabled(...) directly; both modules green locally.

Thiagoag:
Hi @vjuranek can you trigger the CI again?

Thiagoag:
Hi @Naros can you trigger the CI again?

Thiagoag:
Once CI passes and this is merged, would it be possible to include it in the next patch release? Any information on the expected release cadence would be appreciated, as I am planning to pick it up. Happy to help if anything is needed.

vjuranek:
Merged, thanks @Thiagoag ! It will be included in 3.7.0.CR1 which should be out in about one week.

PR #188, opened 2026-09-16
debezium/dbz#2651 Fail the task when the processor group is stopped

### LINKED ISSUE #2651: Task reports RUNNING after the processor group has been stopped
## Bug report

`debezium-connector-cassandra-5`, version 3.6.1.Final, run by a Kafka Connect worker, so the code
path is `AbstractConnectorTask`.

---

```json
{
  "connector.class": "io.debezium.connector.cassandra.Cassandra5Connector",
  "topic.prefix": "cdc",
  "cassandra.config": "/path/to/cassandra.yaml",
  "commit.log.relocation.dir": "/var/lib/debezium/relocation",
  "offset.backing.store.dir": "/var/lib/debezium/offset",
  "commit.log.real.time.processing.enabled": "false"
}
```

---

Apache Cassandra 5.0.8, on-premises on Kubernetes as a StatefulSet, three nodes, RF=2. The
connector runs as a sidecar container on the same pod as Cassandra.

---

When a processor fails, the connector stops its whole processor group, which means CDC has
stopped. The task should not keep reporting `RUNNING` in that state.

---

**What behavior do you see?**

`GET /connectors/<name>/status` reports `RUNNING` for both the connector and the task, forever,
while no CDC events are produced.

Stopping all processors when one fails is intentional and documented on `ProcessorGroup`. The gap
is that the failure is never reported to the task:

```java
catch (Exception e) {
    LOGGER.error("Encountered exception while running {}; stopping all processors.", processor.getName(), e);
    try {
        stopProcessors();
    }
    catch (Exception e2) {
        LOGGER.error("Encountered exceptions while stopping all processors", e2);
    }
}
```

It is raised on an executor thread, so it cannot reach the task thread by itself. And nothing on
the task side looks for it: `AbstractConnectorTask.doPoll()` only drains the queue, and
`getErrorHandler()` returns `Optional.empty()`.

The non-Connect entry point does check. `CassandraConnectorTaskTemplate.run()` polls the group and
throws once it stops:

```java
while (processorGroup.isRunning()) {
    Thread.sleep(1000);
}
throw new CassandraConnectorTaskException("Processor group stopped unexpectedly");
```

So the same invariant is enforced on one entry point and absent on the other.

Any processor failure gets you here. In our case it was the snapshot query failing because one
node was briefly unavailable while the cluster was rolling:

```
io.debezium.DebeziumException: Failed to snapshot table keyvalue in keyspace k1
Caused by: com.datastax.oss.driver.api.core.servererrors.UnavailableException:
    Not enough replicas available for query at consistency QUORUM (2 required but only 1 alive)
```

The node was back within seconds, but CDC stayed stopped for as long as the deployment was
observed.

---

We reproduced it on 3.6.1.Final and have not run a newer release. The three code paths quoted
above are unchanged on the current 3.7.0-SNAPSHOT tree: `getErrorHandler()` still returns
`Optional.empty()`, `doPoll()` still has no check, and the catch block still only logs and calls
`stopProcessors()`.

---

The processors start, one fails, all four are stopped, and the log contains no CDC activity after
that point:

```
INFO  Started Commit Log Post-Processor
INFO  Started Snapshot Processor
INFO  Started Queue Processor [0]
INFO  Started Commit Log Processor
ERROR Encountered exception while running Snapshot Processor; stopping all processors.
      Caused by: UnavailableException: Not enough replicas available for query at consistency
                 QUORUM (2 required but only 1 alive)
INFO  Stopping Commit Log Post-Processor
INFO  Stopping Snapshot Processor
INFO  Stopping Queue Processor [0]
INFO  Stopped Queue Processor [0]
INFO  Stopped Commit Log Post-Processor
INFO  Stopped Commit Log Processor
```

Both the connector and the task report `RUNNING` throughout.

---

**How to reproduce the issue using our tutorial deployment?**

1. Start Cassandra 5 with `cdc_enabled: true`, create a CDC-enabled table and insert some rows.
2. Register the connector with `commit.log.real.time.processing.enabled=false`.
3. Make the snapshot query fail once. Setting `snapshot.consistency=ALL` and making one replica
   unavailable while the connector starts is a reliable way to do it.
4. Confirm `Encountered exception while running Snapshot Processor; stopping all processors.`
   appears in the log.
5. Bring the replica back, so the cluster is healthy again.
6. `GET /connectors/<name>/status` still reports `RUNNING` for the connector and the task, no CDC
   events are produced, `relocation/` stays empty, and files pile up in `cdc_raw/`.

---

## Suggested fix (not part of the official bug report template)

Report the failure through the mechanism the framework already has for it.
`ErrorHandler.setProducerThrowable` forwards to `ChangeEventQueue.producerException`, which
`ChangeEventQueue.poll()` rethrows on the task thread, so `doPoll()` needs no change:

1. build an `ErrorHandler` in `AbstractConnectorTask.start(Configuration)` and return it from
   `getErrorHandler()`
2. pass it to the `ProcessorGroup` and call `setProducerThrowable(e)` in the existing catch block

The handler is optional, so the non-Connect entry point keeps its current behaviour and its own
watchdog. With the default retriable set the failure above is not retriable, so the task moves to
`FAILED` carrying the original cause. Classifying Cassandra's transient driver exceptions as
retriable, by overriding `communicationExceptions()`, would allow in-process retry instead; we
left that out to keep the change minimal.

---

cezarpaulo16:
Fixes debezium/dbz#2651

Stopping every processor when one of them fails is intentional, but the failure was never reported to the embedding task. It is raised on an executor thread, doPoll() only drained the queue, and getErrorHandler() returned an empty Optional, so the task stayed in RUNNING while no CDC events were produced and nothing was able to react to it.

Build an ErrorHandler in the task and report the failure to it from the processor group. ErrorHandler.setProducerThrowable forwards to the ChangeEventQueue, whose poll() rethrows on the task thread, so the task now fails with the original cause attached and doPoll() needs no change. The handler is optional and is not set on the non-Connect entry point, which keeps its existing behaviour and its own watchdog.


cezarpaulo16:
Hello @vjuranek  and @Naros, can you please take a look at this commit?

core/src/test/java/io/debezium/connector/cassandra/CassandraConnectorTaskTest.java
```diff
@@ -73,4 +77,41 @@ public void process() {
         assertFalse(processor2.isRunning());
         assertEquals(0, running.get());
     }
+
+    @Test
+    @Timeout(60)
+    void testProcessorFailureIsReportedToErrorHandler() throws Exception {
```
Naros, line 85:
```suggestion
	@FixFor("debezium/dbz#2651")
    void testProcessorFailureIsReportedToErrorHandler() throws Exception {
```

Naros (commented):
LGTM, just one small nit on the test annotations.

vjuranek (approved):
LGTM, let's wait for the CI.

cezarpaulo16:
Hi @Naros and @vjuranek , I fixed the last commit that failed in CI. Can you please trigger it again?

vjuranek:
I re-triggered the CI, but there seems to be some compilation issue. Could you please take a look @cezarpaulo16 ?

cezarpaulo16:
@vjuranek Yes, sorry for that. I fixed it now. It was the new annotation FixFor that I added; it was missing the import. It builds locally now.

cezarpaulo16:
Hi @Naros  and @vjuranek can you please trigger again?

vjuranek:
Merged, thanks @cezarpaulo16 !

